PR #23, opened 2019-11-21
internal services and service discovery docs

ddollar:


docs/guides/service-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+    $ convox services
+    SERVICE  DOMAIN                                PORTS
+    auth     auth.myapp.0a1b2c3d4e5f.convox.cloud  443:5000
+    web      web.myapp.0a1b2c3d4e5f.convox.cloud   443:3000
+
+The `web` service could reach the `auth` service using `https://auth.myapp.0a1b2c3d4e5f.convox.cloud:443`
+
+> Note that both of these services are available to the public internet.
+
+### Internal Services
```
beastawakens: Should be only `##` rather than `###` ?

docs/guides/service-discovery.md
```diff
@@ -0,0 +1,55 @@
...
+        port: 3000
+
+You would see a `convox services` output similar to this:
+
+    $ convox services
+    SERVICE  DOMAIN                                PORTS
+    auth     auth.myapp.0a1b2c3d4e5f.convox.cloud  443:5000
+    web      web.myapp.0a1b2c3d4e5f.convox.cloud   443:3000
+
+The `web` service could reach the `auth` service using `https://auth.myapp.0a1b2c3d4e5f.convox.cloud:443`
```
beastawakens: Is the `0a1b2c3d4e5f` part pre-known or discoverable? Or does the `web` service actually need to run `convox services` and grep the output to programmatically build the URL for the `auth` service?

camerondgray: Yeah I have basically the same question. We previously used links for this but for an app that is deployed to multiple racks this would be a challenge. If this is the only way to do it for now we should probably add something about setting an env variable with the url so people don't hardcode it.

ddollar: The `0a1b3c3d4e5f.convox.cloud` is the base domain for the entire Rack. I want to revisit how links work in general, and we can update this document at that time. For services within the same app we will likely continue to use `links`. Connections to other apps [on other Racks] would need to be set as configuration in the app.

That being said I don't want to digress into telling people to set environment variables for things here as I think it muddies the topic. We should cover that in a configuration/environment document.

docs/guides/service-discovery.md
ddollar: No, this is a sub-topic of the previous topic.

PR #36, opened 2019-12-04
k8s: better calculation of service-specific environment

ddollar:


ddollar:
[better calculation of service-specific environment](https://trello.com/c/z1yFQL1H/495-better-calculation-of-service-specific-environment)

PR #37, opened 2019-12-04
atom: generate k8s events for state changes

ddollar:


ddollar:
[generate k8s events for atom state changes](https://trello.com/c/7fSamARz/494-generate-k8s-events-for-atom-state-changes)

PR #38, opened 2019-12-04
do: dont use special characters for registry passwords

ddollar:


ddollar:
[dont use special chars for registry passwords](https://trello.com/c/MUjxQQ4B/493-dont-use-special-chars-for-registry-passwords)

PR #39, opened 2019-12-05
gcp: cluster autoscaling

camerondgray:


camerondgray:
[Enable autoscaling by default on GCP](https://trello.com/c/usaWJFq5/497-enable-autoscaling-by-default-on-gcp)

PR #40, opened 2019-12-05
fix for digital ocean authentication

camerondgray:


terraform/cluster/do/main.tf
```diff
@@ -45,6 +45,18 @@ resource "digitalocean_kubernetes_cluster" "rack" {
...
+provider "kubernetes" {
+  version = "~> 1.10"
+
+  alias = "direct"
+
+  load_config_file = false
+
+  cluster_ca_certificate = "${base64decode(digitalocean_kubernetes_cluster.rack.master_auth.0.cluster_ca_certificate)}"
+  host                   = "https://${digitalocean_kubernetes_cluster.rack.endpoint}"
+  token                  = digitalocean_kubernetes_cluster.rack.kube_config[0].master_auth.0.token
```
camerondgray: This is the only bit I am not quite sure about. Is this the correct place to grab the token from?

camerondgray:
[DO installer fails with permissions errors](https://trello.com/c/I6mNXprB/481-do-installer-fails-with-permissions-errors)

PR #42, opened 2019-12-05
do: restore token auth

ddollar:


ddollar:
[DO installer fails with permissions errors](https://trello.com/c/I6mNXprB/481-do-installer-fails-with-permissions-errors)

PR #43, opened 2019-12-05
aws: switch to managed node groups

ddollar:


ddollar:
[use managed node groups for eks](https://trello.com/c/D5b9IPi7/501-use-managed-node-groups-for-eks)

terraform/cluster/aws/main.tf
```diff
@@ -66,56 +27,35 @@ resource "aws_eks_cluster" "cluster" {
...
   ]
 
-  filename = pathexpand("~/.kube/config.aws.${var.name}")
-  content = templatefile("${path.module}/kubeconfig.tpl", {
-    ca       = aws_eks_cluster.cluster.certificate_authority.0.data
-    cluster  = aws_eks_cluster.cluster.id
-    endpoint = aws_eks_cluster.cluster.endpoint
-  })
-}
+  count = 3
```
camerondgray: I couldn't find a reference to the count argument in the docs for this. For the GCP stuff you need to you use initial_count if you enable autoscaling but that didn't seem to be a argument option for this either:  https://www.terraform.io/docs/providers/aws/r/eks_node_group.html . Am I just looking in the wrong place?

ddollar: `count` is a terraform thing. it means create n of these

camerondgray: got it. Thanks for clarifying.

PR #49, opened 2019-12-10
terraform: move authentication out to env vars

ddollar:


install/gcp/README.md
```diff
@@ -12,7 +12,15 @@
...
 - Click **Create**
-- Download the credential file to `~/.config/gcloud/terraform.json`
+- Download the credential file
+
+## Configuration
+
+### Environment Variables
+
+- `GOOGLE_CREDENTIALS` (path or contents of the credentials file)
+- `GOOGLE_PROJECT` (required)
```
camerondgray: It might be worth noting that it is the ID of the project and not the name. I have made this mistake before and it looks like you just have some permissions problem when you try to apply the template

PR #51, opened 2019-12-11
reference docs

ddollar:


ddollar:
Pulled `instances ssh` back in since this CLI needs to be backwards-compatible but there will be no keys/SSH configured/available for k8s racks

PR #107, opened 2020-01-31
Expose the 'disk_size' param of module 'aws_eks_node_group'.

jfmyers:
This PR exposes the `disk_size ` parameter of the `aws_eks_node_group` module up to the `system/aws` template as the variable `node_disk`.

Example usage:
```
module "system" {
  source = "../../terraform/system/aws"

  node_disk = var.node_disk
  ...
}
```


ddollar:
Thanks! Please note that if you're upgrading to the latest master on this repo the Terraform installation bits have changed quite a bit. There are no longer templates in the `install/` subdirectory so if you're using those you'll want to copy it out somewhere before pulling master.

The new recommended method for installation involves grabbing the latest CLI from a release of this repo and using `convox rack install` which will wrap the Terraform bits and give you local rack management via`convox racks`

See: https://github.com/convox/convox/tree/a3b736525967365c952ac2a3b5abf7e68d006e41/docs/installation/production-rack

jfmyers:
No problem. Thanks for the heads up. Quick question on that. If I want to use `convox rack install`(as opposed to terraform config) and customize the params such as the `node_instance` or `node_disk`, what's the best way to do that via the cli?

ddollar:
`convox rack install aws <name> node_disk=30`

PR #145, opened 2020-02-18
Fix install rack documentation

scottmessinger:
Add --name=<name>


scottmessinger:
My fault -- this is how it's done in gen.2.

PR #146, opened 2020-02-19
docs: ca trust command backwards compatibility

rabidaudio:
On macos, `base64` decode is `-D` not `-d`. Just changed it to cross-platform and more clear `--decode`.

### DIFF
--- a/docs/installation/development-rack/macos.md
+++ b/docs/installation/development-rack/macos.md
@@ -42,5 +42,5 @@ you can trust the Rack's CA certificate.
 
 This certificate is generated on your local machine and is unique to your Rack.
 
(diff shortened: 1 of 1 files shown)


ddollar:
Which version of OSX are you on? On Catalina I see support for both `-D` and -d`

rabidaudio:
I guess I'm still on Mojave. They must have made that change with Catalina.

Screen Shot 2020-02-19 at 11 30 04 AM

```
$ echo "aGVsbG8K" | base64 -d
base64: invalid option -- d
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
$ echo "aGVsbG8K" | base64 -D
hello
```

PR #161, opened 2020-02-25
docs: fix broken links

camerondgray:


.gitignore
```diff
@@ -3,3 +3,5 @@
 coverage.txt
 terraform.tfvars
 terraform/platform/convox.platform
+terraform.tfstate
```
ddollar: These files should not be in your repo any more, you can delete them

PR #167, opened 2020-02-26
Update timer.md

erichummel:
The schedule format changed from v2 to v3 (and caused me some grief until I figured it out). This change makes the format consistent in both examples


ddollar:
Thaks!

PR #193, opened 2020-03-19
docs: Information on dealing with v2 Racks

beastawakens:


docs/help/changes.md
```diff
@@ -2,6 +2,10 @@
 
 This document outlines the changes from Version 2 (date-based version) Racks to Version 3.x Racks.
 
+## Overview of Rack differences
+
+![](rack-versions.png)
```
ddollar: Why make this an image instead of text (e.g. markdown table)

docs/help/version2-racks.md
```diff
@@ -0,0 +1,48 @@
+# Version 2 Racks
```
ddollar: I don't feel like this should be in the main help docs. Maybe in the v2 docs site and/or Console? 

I want to avoid as much confusion as possible for new folks coming in that have never heard of v2.

docs/help/changes.md
beastawakens: 😄 Good point!

docs/help/version2-racks.md
ddollar: Or can this information be included inline like the rest of the stuff in this doc?

ddollar:
Seems like this is covered by #198, can it be closed?

PR #202, opened 2020-03-26
docs: changing links to convox.yml

beastawakens:


camerondgray:
These changes seem fine but don't you need to actually rename the convox.yml.md file as well?

beastawakens:
> These changes seem fine but don't you need to actually rename the convox.yml.md file as well?

The page generator already changes the file name in the url to match, but it is probably a good idea to change it as well... 👍

PR #207, opened 2020-04-01
docs: initial example apps

beastawakens:


docs/example-apps/django.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. Requirements.txt is copied and `pip install` is run. This happens first because it is slow and something that's done infrequently. After running once, this step will be cached unless the cache is busted by later edits to `requirements.txt`.
+
+2. The application source is copied over. These files will change frequently, so this step of the build will very rarely be cached.
+
+#### convox.yml
+
+The [convox.yml](https://github.com/convox-examples/django/blob/master/convox.yml) file explains how to run the application. This file has two sections.
+
+1. Resources: These are network-attached dependencies of your application. In this case we have a single resource which is a postgres database. When [running locally](https://docs.convox.com/development/running-locally) Convox will automatically startup up a container running Postgres and will inject a ```DATABASE_URL``` environment variable into your application container that points to the Postgres database. When your application is [deployed](https://docs.convox.com/deployment/deploying-to-convox) to production Convox will startup an RDS postgres database for your application to use.
```
camerondgray, line 44: I would change this to when deployed instead of [running locally] and then delete everything after "...that points to the Postgres database" since we don't do the RDS thing on v3

docs/example-apps/README.md
```diff
@@ -0,0 +1,5 @@
+# Example Apps
+
+All of our example apps can be found in [this Github repo](https://github.com/convox-examples/).  Over time we will add to them with further languages and frameworks.  We welcome simple submissions of further projects or requests for specific examples.
```
ddollar, line 3: Github organization (not repo)

PR #212, opened 2020-04-06
docs: adding in extra cli commands

beastawakens:


docs/reference/cli/apps.md
```diff
@@ -117,6 +117,50 @@ Enable termination protection
     $ convox apps lock
     Locking myapp... OK
 
+## apps params
```
ddollar, line 120: Let's just leave app and rack params out of docs until we use them, feel like this will be confusing for not much gain.

docs/reference/cli/login.md
```diff
@@ -0,0 +1,17 @@
+# login
+
+## login
+
+Authenticate your CLI with a Rack or Console installation
```
ddollar, line 5: login is really just for Console now

direct racks should use RACK_URL

docs/reference/cli/rack.md
```diff
@@ -53,6 +51,58 @@ Get logs for the rack
     2020-02-10T13:38:04Z service/web/a55eb25e-90f5-4301-99fd-e35c91128592 id=f492a0dce931 ns=api at=SystemGet method="GET" path="/system" response=200 elapsed=332.219
     ...
 
+## rack mv
+
+Move locally installed Racks to be managed by a organization within a Console installation, or move an organization's Rack back down to be managed locally.
```
ddollar, line 56: Is there a way to say this more simply that still captures the subtlety?

"Locally installed" sounds like the Rack is running on my machine

"Console installation" is a bit overloaded

Is there a single short sentence that describes the effect of the change?

PR #216, opened 2020-05-05
add `--name` flag to `convox rack install local` development installation tutorial

krmannix:
I looks like the CLI for `convo rack install` was updated, so that the `dev` argument needs to be specified with `--name`


ddollar:
Which version of the CLI are you running? The correct command is `convox rack install local dev`

If you're on the v2 CLI, make sure you get the latest by following the instructions in the docs.

krmannix:
`convox version` gives me `client: 20200428134512`

PR #237, opened 2020-05-25
Update introduction.md

conarro:
Pass the app name as an argument to `apps create` instead of via the `--name` flag (which results in `unknown flag: --name`)


ddollar:
Thank you!

PR #238, opened 2020-05-25
Add to docs on logging without copying from web UI

bigshika:
I ran into an issue where I'm setting up the Convox CLI on a new machine, however I already have workflows set up, so the existing login screenshot is inaccurate to my situation. In this case, I just need to know the CLI command.

image


ddollar:
Thank you!

PR #239, opened 2020-06-02
aws: fix eks perm for elb

camerondgray:
@ddollar fixes the issue described here: https://medium.com/faun/aws-eks-the-role-is-not-authorized-to-perform-ec2-describeaccountattributes-error-1c6474781b84 which has been reported a few times by people with fresh AWS accounts

### DIFF
--- a/terraform/cluster/aws/iam.tf
+++ b/terraform/cluster/aws/iam.tf
@@ -35,10 +35,6 @@ data "aws_iam_policy_document" "assume_eks" {
   }
 }
 
-data "aws_iam_policy_document" "assume_elb" {
-
-}
-
 resource "aws_iam_role" "cluster" {
   assume_role_policy = data.aws_iam_policy_document.assume_eks.json
   name               = "${var.name}-cluster"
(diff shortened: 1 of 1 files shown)


ddollar:
the fix in this commit is not the same suggested in the blog post, has this been tested on a fresh aws account?

camerondgray:
The blog post references a few options. First you can create an ELB with the account from the console or CLI which then gives the account the required permissions. Second ```$ aws iam create-service-linked-role --aws-service-name "elasticloadbalancing.amazonaws.com"```
Which is what I believe I am doing here. The third option is to directly give the role ec2.describeinstances permissions. I felt like this option was the best because it seemed clean and if there was additional permissions now or in the future around loadbalancing that giving the ELB role and all its associated perms would solve for them. The complete list of permissions for the role can be found here: https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/elb-service-linked-roles.html. I have not yet tested this on a new account. I wanted to get your eyes on it first.

camerondgray:
@ddollar After much testing I have updated the fix. This fix applies the minimum permissions required and works for rack creating and deletion. I have replicated the original error with a new AWS account and verified the fix with the same account

PR #245, opened 2020-06-10
aws: add govcloud support

camerondgray:


ddollar:
Looking through the docs it seems like this is probably actually what you should use for this:

https://www.terraform.io/docs/providers/aws/d/partition.html

camerondgray:
modified with the notes from @ddollar thanks! Re-tested for Rack creation and deletion on both govcloud and non-govcloud regions

PR #253, opened 2020-06-25
k8s: return ports in process list

beastawakens:


provider/k8s/process.go
```diff
@@ -567,6 +567,11 @@ func (p *Provider) processFromPod(pd ac.Pod) (*structs.Process, error) {
 		}
 	}
 
+	ports := []string{}
+	for _, p := range cs[0].Ports {
+		ports = append(ports, fmt.Sprintf("%d:%d", &p.HostPort, &p.ContainerPort))
```
ddollar, line 572: should not need the pointer references here

PR #255, opened 2020-06-30
docs: fixed incorrect reference to "memcached"

azazeal:
The memcached literal was incorrectly carried over into the Postgres resource documentation. This PR fixes this.


ddollar:
Thanks!

PR #257, opened 2020-07-01
k8s: return ports in process list

beastawakens:


provider/k8s/process.go
```diff
@@ -569,14 +569,18 @@ func (p *Provider) processFromPod(pd ac.Pod) (*structs.Process, error) {
 
 	ports := []string{}
 	for _, p := range cs[0].Ports {
-		ports = append(ports, fmt.Sprintf("%d:%d", p.HostPort, p.ContainerPort))
+		if p.HostPort == 0 {
```
ddollar, line 572: Make sure to add a test for this case

provider/k8s/process_test.go
```diff
@@ -6,7 +6,6 @@ import (
 
 	"github.com/convox/convox/pkg/structs"
 	"github.com/convox/convox/provider/k8s"
-
```
ddollar, line 9: This likely violates gofmt

provider/k8s/process.go
beastawakens: Yep, there is, the 3rd one in the list!

beastawakens:
> * make sure to run `gofmt` on all files you edit.. probably want to do this automatically on save
> * structure your commits to be fairly atomic with descriptive commit messages, avoid "WIP"

- I swore I had code set up to do that... 🤔 I'll check it.
- Yeah, that crept in from rebasing from the previous, now defunct, branch 😛

beastawakens:
@ddollar All sorted now!

PR #259, opened 2020-07-02
aws: pass through source ip information in custom balancers

jsierles:
Currently there's no way to get the client IP for a custom balancer. [Docs here suggest](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol) that the NLB will pass the correct source IP when using instances as targets, unlike the currently-used classic ELB which would require enabling the proxy protocol.

> If you specify targets by instance ID, the source IP addresses provided to your applications are the client IP addresses. However, if you prefer, you can enable proxy protocol and get the client IP addresses from the proxy protocol header.

### DIFF
--- a/provider/k8s/template/app/balancer.yml.tmpl
+++ b/provider/k8s/template/app/balancer.yml.tmpl
@@ -5,6 +5,7 @@ metadata:
   name: balancer-{{.Balancer.Name}}
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
   labels:
     balancer: {{.Balancer.Name}}
     service: {{.Balancer.Service}}


ddollar:
Couple of issues here that I can see:

1) I'm fairly sure Kubernetes-created NLBs use IP mode and not Instance mode and thus would still need proxy protocol to be able to pass the source IP. This seems tough to explain/reconcile.

2) This change would potentially recreate the custom balancers for any existing users which would change their hostnames and break their production apps

jsierles:
In this screenshot, you can see the NLB created as the Convox router service is using instance-based targets. I have not tested yet, but will check to see which IP is received by nginx. https://d.pr/free/i/iLYGde

Related: I enabled the proxy protocol for my classic ELB balancer and setup nginx to accept it. This works and allowed me to append the real IP to `X-Forwarded-For`. This suggests that you could use either approach for either LB.

I see what you mean about recreating someone's service. What about adding an option to `convox.yml` that would trigger NLB mode and/or enable the proxy protocol?

ddollar:
Did some testing and changing the load balancer annotation to NLB leaves the old load balancer in place until you manually delete it.

jsierles:
Cool! Does that mean you feel it's safe to upgrade Convox balancers to NLB? Also, the recent PR about 'external_traffic_policy' - does that pass the source IP directly without proxy protocol?
GitHub <notifications@github.com> wrote:
“Did some testing and changing the load balancer annotation to NLB leaves the old load balancer in place until you manually delete it.”

ddollar:
Yes and yes

jsierles:
Very nice. Last minute save!
GitHub <notifications@github.com> wrote:
“Yes and yes”

jsierles:
One more tricky thing to throw in here. We will be using Cloudflare Spectrum in front of this balancer for DDoS protection. It will only pass client IP via the proxy protocol. I've heard - but didn't test yet - that ELBs will accept the proxy protocol. Do you think it will do the right thing here and pass the client IP on either as the source IP, or via X-Forwarded-For? I'm going to test with spectrum and a manually-launched NLB. But thought I'd mention this too.
Joshua Sierles <joshua@hey.com> wrote:
“Very nice. Last minute save!

GitHub <notifications@github.com> wrote:“Yes and yes”

PR #270, opened 2020-07-15
AWS is using six required fields in cron

adrianoesch:
"Cron expressions have six required fields, which are separated by white space." https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html#CronExpressions

### DIFF
--- a/docs/reference/primitives/app/timer.md
+++ b/docs/reference/primitives/app/timer.md
@@ -28,12 +28,12 @@ A Timer is defined in [`convox.yml`](../../../configuration/convox-yml.md).
 Cron expressions use the following format. All times are UTC.
 
 ```
-.----------------- minute (0 - 59)
-|  .-------------- hour (0 - 23)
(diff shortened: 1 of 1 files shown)


ddollar:
Thank you for your pull request but the convox/convox repo uses Kubernetes CronJob implementation which uses a different schedule format.

adrianoesch:
Ok, i tried with 5 and ran into timer registration errors. Also, the convox.yml [docs](https://docs.convox.com/configuration/convox-yml#timers) use 6 too.

ddollar:
Are you on a v2 (ECS) or v3 (EKS) Rack?

adrianoesch:
Using the client (v3.0.15) i get `Version   20200420094617` when inspecting via `convox rack`. I guess we are on some kind of enterprise solution. Anyhow, it's not really an issue for me anymore, so you probably know better ;)

PR #283, opened 2020-07-21
k8s: use docker buildkit for remote build cache

ddollar:
Uses Docker's buildkit to enable a build cache stored in the registry.

The builder then uses this stored build cache to allow for cached build without needing to pin to a specific instance.


ddollar:
Test with the `buildkit` release

ddollar:
Closing for now

ddollar:
Needs a newer Docker available at the infrastructure providers to use Buildkit

PR #296, opened 2020-08-10
docs: Whitelist param, kubeconfig setup, deploy keys

beastawakens:


docs/management/direct-k8s-access.md
```diff
@@ -0,0 +1,25 @@
+# Direct Kubernetes Access
+
+Convox makes managing and operating your Kubernetes cluster extremely easy and simple by abstracting a lot of the unnecessary complexities away  We know, however, that there may be times that you will want to dive deeper and access the k8s primitives and resources yourself.
```
ddollar, line 3: Missing a period. Try not to use shortening jargon like `k8s` in documentation.

docs/management/direct-k8s-access.md
```diff
@@ -0,0 +1,25 @@
...
+
+This could be deep debugging of an edge-case networking issue, examination of an underlying cloud infrastructure issue, gaining a deeper understanding of how everything ties together, or simply to install a Kubernetes native component or plugin directly.
```
ddollar, line 5: This idea should be part of the previous paragraph.

docs/management/direct-k8s-access.md
```diff
@@ -0,0 +1,25 @@
...
+
+Convox provides an API proxy to Kubernetes that runs on your Rack which allows you to grant access to the underlying system, whilst still respecting and adhering to the Role-Based Access Control (RBAC) that Kubernetes provides.  This is actually a lot easier and more manageable than providing direct Kubernetes credentials to the developers in your team!
```
ddollar, line 7: We don't really do anything with Kubernetes RBAC. The main advantage is because of this proxy if you're using Console you can delegate per-user access to Kubernetes.

docs/management/direct-k8s-access.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Configure kubectl to Point at Your Rack
+
+Convox allows you to securely connect your `kubectl` to your Convox created Kubernetes cluster by exporting a [kubeconfig](../../reference/cli/rack#rack-kubeconfig) that will connect you to a Kubernetes API Proxy running inside your Rack. This allows you to use `kubectl` without directly exposing the credentials for your Kubernetes cluster. For example if your Rack is called `myrack` you could point your local `kubectl` to your Rack cluster as follows
+
+```sh
+$ convox switch myrack
+$ convox rack kubeconfig > /tmp/myrack-config
+$ export KUBECONFIG=/tmp/myrack-config
```
ddollar, line 16: For a real example we should probably have them put this file in `~/.kube` so it doesn't get autodeleted.

docs/installation/production-rack/aws.md
```diff
@@ -37,11 +37,12 @@ The following environment variables are required:
...
\ No newline at end of file
+| Name        | Default       | Description                                                             |
+| ----------- | ------------- | ----------------------------------------------------------------------- |
+| `cidr`      | `10.1.0.0/16` | CIDR range for VPC                                                      |
+| `node_disk` | `20`          | Node disk size in GB                                                    |
+| `node_type` | `t3.small`    | Node instance type                                                      |
+| `private`   | `true`        | Put nodes in private subnets behind NAT gateways                        |
+| `region`    | `us-east-1`   | AWS Region                                                              |
+| `syslog`    |               | Forward logs to a syslog endpoint (e.g. `tcp+tls://example.org:1234`)   |
+| `whitelist` | `0.0.0.0/0`   | Comma delimited list of CIDR Ranges that are allowed to access the Rack |
```
camerondgray, line 48: I am wondering if we want to leave whitelist out of this list because it's really only for private console customers.

docs/management/deploy-keys.md
```diff
@@ -0,0 +1,39 @@
+# Deploy Keys
+
+Deploy keys are limited scope API keys that allow you to run some limited commands from a remote environment (such as CI) without needing to store/use/expose your user credentials.
```
camerondgray, line 3: might expand on (such as CI) with some more verbose like "such as continuous integration system like Jenkins or Travis"

docs/management/direct-k8s-access.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+This will export the proxy configuration to a temporary file and then point your local `kubectl` environment at that location so you can connect to your Rack's cluster. You will need to perform this step before you can execute any `kubectl` commands against your cluster.
+
+> By default, `kubectl` looks for a file named config in the `$HOME/.kube` directory. You can specify other kubeconfig files by setting the `KUBECONFIG` environment variable or by setting the `--kubeconfig` flag.
+
+In this way, you can produce a kubeconfig file for all the Racks you require and simply change which file your `kubectl` command refers to to change which Rack it talks to.
+
+If you remove a user's access to a Rack in Convox, then they will also lose access to the underlying Kubernetes infrastructure, which is important from a security point of view.
```
camerondgray, line 25: I would replace "If you remove a user's access to a Rack in Convox" to "if you remove a users's access to your Convox organization"

docs/installation/production-rack/aws.md
beastawakens: Or a non-Console Rack it could be useful for as well...

PR #310, opened 2020-10-13
docs: missed options

hlopetz:
i spent some time realizing how to specify an alternative yml manifest file. there is such option, but it is not documented. after a number of failures i found `-m` works, so i want to share my solution with others.

feel free to add the missing parts.


jasontclark (approved):
LGTM. :+1:

PR #311, opened 2020-10-15
rack: Skip invalid parameters on import

beastawakens:
Exporting an App from v2 and importing to v3 will cause errors due to the lack of matching (or any) app params on v3.  Instead we can skip over them during Import.

### DIFF
--- a/provider/k8s/app.go
+++ b/provider/k8s/app.go
@@ -281,7 +281,7 @@ func (p *Provider) appParametersUpdate(a *structs.App, params map[string]string)
 		}
 	}
 
-	fmt.Print("Skipping invalid parameters: %s ...", strings.Join(invalidParameters, ", "))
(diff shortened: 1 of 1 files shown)


provider/k8s/app.go
```diff
@@ -272,14 +272,17 @@ func (p *Provider) appNameValidate(name string) error {
...
-			return errors.WithStack(fmt.Errorf("invalid parameter: %s", k))
+			invalidParameters = append(invalidParameters, k)
+		} else {
+			a.Parameters[k] = v
 		}
-
-		a.Parameters[k] = v
 	}
 
+	fmt.Printf("Skipping invalid parameters: %s ...", strings.Join(invalidParameters, ", "))
```
ddollar, line 284: This message is a little scary. Maybe just remove the print entirely since there are no parameters on v3? If you keep it, make it something other than "invalid" which sounds like they did something wrong.

ddollar (approved):

PR #312, opened 2020-10-29
k8s: nginx timeout config

beastawakens:


beastawakens:
@ddollar all tested and it works.  The annotation which makes the difference when the backend service takes longer than 60s to respond is the `nginx.ingress.kubernetes.io/proxy-read-timeout` one.  I'm thinking we should maybe keep the other annotations in there, to cover other timeout situations (could make the `timeout` setting in convox.yml more flexible in the future possibly), but also happy to drop the other annotations if you just want to leave it with the proxy read one.

ddollar (approved):

PR #315, opened 2020-11-18
rack: allow annotations to populate serviceaccount's and deployments

beastawakens:


provider/k8s/template/app/service.yml.tmpl
```diff
@@ -8,10 +8,23 @@ metadata:
...
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    {{ range keyValue .Annotations }}
+    {{.Key}}: "{{ quoteEscape .Value}}"
+    {{ end }}
+  namespace: {{.Namespace}}
+  name: serviceaccount-{{.Service.Name}}
```
ddollar, line 23: Seems like the name could just be the service name since it will already be namespaced by "serviceaccount", e.g.

`kubectl get serviceaccount/web2`

ddollar (changes requested):

ddollar (approved):

PR #316, opened 2020-11-24
Update provider to ~>3.0

mrobinson513:


mrobinson513:
Opened here by accident, sorry fellas

PR #326, opened 2021-03-01
aws: enable spot instance types

ddollar:
Adds a new parameter to AWS racks: `spot_types`

Set this parameter to one or more instance types to enable spot instances in your Rack.

The Rack will still run a minimum of 3 on demand instances (one in each AZ) to keep a bare minimum of infrastructure running should all spots disappear.

Note: Because AWS managed node groups currently require a minimum of one instance setting the `spot_types` parameter will cause your instance to have a minimum of 6 instances (3 on-demand, 3 spot)


ddollar:
Needs a resolution to kubernetes/autoscaler#3871 in order to function correctly

PR #333, opened 2021-03-24
k8s: Only import service specific env vars

beastawakens:


ddollar (commented):
Probably better to make one function that resolves the whole environment for a given service rather than putting so much logic in the template.

beastawakens:
@ddollar I think this is the final form now.  Good to merge?

ddoc:
Implementing this now would break expected functionality for a lot of customers. I propose updating docs to reflect current behavior and updating this PR to enable more strict behavior only with a rack parameter.

slaskis:
any plans to release this?

nightfury1204:
very old pr

PR #338, opened 2021-04-19
rack: only auto-update within the same major/minor version

beastawakens:


heronrs (changes requested):
Code wise it looks good, but I think we have two problems that still remain.

This PR only accounts for a cli update, correct?
I think we need it for rack updates too.

As for the logic, we want to prevent users from skipping minor/major releases.
E.g say `v3.1.3` is on `k8s v1.18` and we try updating to `v3.3.0` which has `k8s v1.20`. We want to prevent users from doing that so users update it incrementally, going to `v3.2.X` and then `v3.3.0` (k8s itself only updates by 1 minor, incrementally)

As for always using the latest minor version I think it makes sense, but we still need to prevent skipping minor/major versions.

heronrs (approved):
LGTM!

PR #339, opened 2021-05-05
terraform: updates for arm64

ddollar:


ddollar:
Need to figure out multiplatform docker builds in the Github Actions before this can be merged.

slaskis:
@ddollar I'm sure you're aware but when trying to use convox on my m1 I keep coming back to this PR so I did a little research and this seems like the currently recommended way for multi platform docker builds in github actions:

https://github.com/docker/setup-buildx-action

Is there something else missing that I don't see? If you need, I would be happy to create a PR to test this out.

PR #347, opened 2021-06-23
docs: update example-apps rails

kaiomagalhaes:


docs/example-apps/rails.md
```diff
@@ -1,45 +1,30 @@
...
-#### [convox.yml](https://github.com/convox-examples/rails/blob/master/convox.yml)
+### Preparing your application
 
-The `convox.yml` manifest explains how to run the application. The manifest for this application has two sections:
-
-1. Resources: These are network-attached dependencies of the application. In this application we have a single resource, a `postgres` database.
+To deploy your rails application in your local rack your need to make a few changes to it. 
 
-2. Services: These are the web-facing services of the application. This application has a single service named `web` which is built from the local directory.
+1. To run your application you will need a [Dockerfile]. Create a file in the root of your project with the name `Dockerfile` and the content of our [example](https://github.com/convox-examples/rails). Notice that ours is using ruby-3.0, if you need a different version feel free to change the version on the first line of the file.
```
beastawakens, line 16: Image references should be valid references from Docker Hub if they change the version.
Maybe if you're going to remove the section describing what the Dockerfile is doing, then link to Dockerfile docs for people to understand that themselves?

docs/example-apps/rails.md
```diff
@@ -1,45 +1,30 @@
...
+2. To tell Convox how it should understand your application, your project needs to have a [convox.yml](https://docs.convox.com/configuration/convox-yml). Create a file in the root of your project with the name `convox.yml` and the content of our [example](https://github.com/convox-examples/rails/blob/master/convox.yml).
```
beastawakens, line 17: We really more in here to describe what the convox.yml is doing and what options people can have.

docs/example-apps/rails.md
```diff
@@ -1,45 +1,30 @@
...
+3. To avoid Docker from loading unnecessary files you need to define a [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file). Create a file in the root of your project with the name `.dockerignore` and the content of our [example](https://github.com/convox-examples/rails/blob/master/.dockerignore).
+4. If you are using Rails 6.0+ you need to add the following line inside `config/environments/development.rb` ```config.hosts << 'api.roadrunner.dev.convox'```
```
beastawakens, line 19: Why do we need this?
(as in, we should explain why they need to add this line)

docs/example-apps/rails.md
```diff
@@ -1,45 +1,30 @@
...
+5. In your gemfile add the following line `gem 'pg'`
+6. To use Convox' database resource, in your `config/database.yml` add the following line under `default` ```  url: <%= ENV['DATABASE_URL'] %>```. Your `database.yml` should look like [this](https://github.com/convox-examples/rails/blob/master/config/database.yml)
```
beastawakens, line 21: We should really still reference where the env var comes from and why it's important.  If the customer has a different name for their resource in the convox.yml then this line won't be true.

docs/example-apps/rails.md
```diff
@@ -71,8 +56,42 @@ And you can deploy your application to production
...
+```
+convox env set SECRET_KEY_BASE="$(rails secret)"
+```
+
 Finally you can retrieve the URL from your production application with
 
 ```bash
 convox services
 ```
+
```
beastawakens, line 70: Hmm, if this is a step-by-step guide, I'm not convinced that leaving the explanations to the end helps the end-user understand what's happening 🤔

beastawakens (changes requested):

docs/example-apps/rails.md
```diff
@@ -71,8 +56,42 @@ And you can deploy your application to production
...
+### Custom Application Components
+
+#### [Dockerfile](https://github.com/convox-examples/rails/blob/master/Dockerfile)
+
+Starting from the [ruby:3.0.0](https://hub.docker.com/_/ruby/) image, the `Dockerfile` defines the steps necessary to turn the application code into an image that is ready to run. 
+
+This `Dockerfile` has 3 steps and they are executed in a particular order to take advantage of Docker's build caching behavior.
+
+1. `bundle install` and `yarn install` are run to install dependencies after copying just the files needed to run these commands. This will ensure that the output these commands are cached unless one of these files changes.
```
jasontclark, line 79: > This will ensure that the output these commands are cached unless one of these files changes.

Should that read, "This will ensure that the output **of** these commands is cached unless one of these files changes."?

docs/example-apps/rails.md
```diff
@@ -71,8 +56,42 @@ And you can deploy your application to production
...
+
+2. The application source is copied over. These files will change frequently so this step of the build will very rarely be cached.
+
+3. Finally, after setting the appropriate environment variables the assets are precompiled.
```
jasontclark, line 83: > Finally, after setting the appropriate environment variables the assets are precompiled.

I'd move the comma from "Finally," to "variables,".

docs/example-apps/rails.md
```diff
@@ -71,8 +56,42 @@ And you can deploy your application to production
...
+
+#### [convox.yml](https://github.com/convox-examples/rails/blob/master/convox.yml)
```
jasontclark, line 85: I think I'd have to agree w/ @beastawakens on moving this explanation of what the convox.yml file is up closer to step 2.

jasontclark (changes requested):
Please have a look at my review comments.  Thanks!

docs/example-apps/rails.md
kaiomagalhaes: @beastawakens my goal here is to provide a docker image that they can use out of the box. We are using an official ruby imager from Dockerhub. I'll add the reference for our Dockerfile docs

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
kaiomagalhaes: 👍🏻

docs/example-apps/rails.md
beastawakens: I guess my point is that if we're assuming the user doesn't know what a Dockerfile is, or how to Dockerise their application, we should be clear with instructions like changing base image version.  Otherwise they may change the version number, and have everything fail straight away because that image doesn't exist in Docker Hub for whatever reason.

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
 
-Once you are all setup you can switch to your local rack with ```convox switch [rack name]``` and start your local application with ```convox start``` (make sure you are in the root directory).
-
-You should now be able to access your application by going to [https://web.rails.convox](https://web.rails.convox). If you renamed anything you may need to modify your local URL. The format is https://[service name].[app name].convox
-
-### Custom Application Components
+### Preparing your application
 
-#### [Dockerfile](https://github.com/convox-examples/rails/blob/master/Dockerfile)
+To deploy your rails application in your local rack your need to make a few changes to it.
```
beastawakens, line 14:
```suggestion
Whilst our example Rails app will run fine straight away, to deploy your own rails application in your local rack you will need to make a few changes to it. 
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+    build: .
+    port: 3000 # Here we are opening the port in which our applicatin will run. If we need a different port we change it here
+    environment: # Here we define the environment variables that we will use in our application. 
+      - SECRET_KEY_BASE=
+    resources: # Here we are linking our service with the resource we want to use. Since our resource is called "database", Convox will create a DATABASE_URL that can be read by our application.
+      - database
+```
+
+**3.** To avoid Docker from loading unnecessary files you need to define a [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file). Create a file in the root of your project with the name `.dockerignore` and the content of our [example](https://github.com/convox-examples/rails/blob/master/.dockerignore).
+**4.** On Rails 6.0+ you need to define any host you will use to access your application. Since Convox generates a host that is different from `localhost` and `0.0.0.0` you need to define it in your application. More information about this configuration [here](https://guides.rubyonrails.org/configuring.html#configuring-middleware).
```
beastawakens, line 56: Needs an extra line space, otherwise the list just runs on.

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+[This](https://github.com/convox-examples/rails/blob/master/Dockerfile) `Dockerfile` has 3 steps and they are executed in a particular order to take advantage of Docker's build caching behavior.
 
 1. `bundle install` and `yarn install` are run to install dependencies after copying just the files needed to run these commands. This will ensure that the output these commands are cached unless one of these files changes.
 
 2. The application source is copied over. These files will change frequently so this step of the build will very rarely be cached.
 
 3. Finally, after setting the appropriate environment variables the assets are precompiled.
 
-#### [convox.yml](https://github.com/convox-examples/rails/blob/master/convox.yml)
+To run your application you will need a [Dockerfile]. Create a file in the root of your project with the name `Dockerfile` and the content of our [example](https://github.com/convox-examples/rails). Notice that ours is using ruby-3.0, if you need a different version feel free to change the version on the first line of the file.
```
beastawakens, line 26: Is `[Dockerfile]` meant to link to somewhere?

beastawakens:
```suggestion
To run your application you will need a [Dockerfile]. Create a file in the root of your project with the name `Dockerfile` and the content of our [example](https://github.com/convox-examples/rails). Notice that ours uses `ruby-3.0` as a base image, if you need a different version feel free to change the version on the first line of the file to a valid ruby Docker image.
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+  web:
+    build: .
+    port: 3000 # Here we are opening the port in which our applicatin will run. If we need a different port we change it here
+    environment: # Here we define the environment variables that we will use in our application. 
+      - SECRET_KEY_BASE=
+    resources: # Here we are linking our service with the resource we want to use. Since our resource is called "database", Convox will create a DATABASE_URL that can be read by our application.
+      - database
+```
+
+**3.** To avoid Docker from loading unnecessary files you need to define a [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file). Create a file in the root of your project with the name `.dockerignore` and the content of our [example](https://github.com/convox-examples/rails/blob/master/.dockerignore).
```
beastawakens, line 55:
```suggestion
**3.** To stop Docker from loading unnecessary files you should define a [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file) file. Create a file in the root of your project with the name `.dockerignore` and the content of our [example](https://github.com/convox-examples/rails/blob/master/.dockerignore).
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+**4.** On Rails 6.0+ you need to define any host you will use to access your application. Since Convox generates a host that is different from `localhost` and `0.0.0.0` you need to define it in your application. More information about this configuration [here](https://guides.rubyonrails.org/configuring.html#configuring-middleware).
+
+The URL that Convox generates follows the following format:
```
beastawakens, line 58:
```suggestion
The URL that Convox automatically generates on your local Rack follows the following format:
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+
+```https://[service name].[app name].convox```
+
+In our convox.yml the service name is `web`, thus your URL will be something like:
```
beastawakens, line 62:
```suggestion
In our convox.yml the service name is `web`, thus your URL will be:
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+
+```https://web.[app name].convox```
+
+Where if you run Convox from the project's application then `app name` will be the folder name. Notice that you can also pass the app name as a parameter i.e: `-a my-app`, in which the host will be:
```
beastawakens, line 66:
```suggestion
By default, the name of your `app` is taken to be the name of the directory you are in.  You can use a different app name by adding the `--app/-a` flags to any Convox command.
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
 
-This file is configured to read database credentials from the `DATABASE_URL` environment variable.
+`https://web.my-app.convox`
```
beastawakens, line 68: Don't need an extra example if the wording above is good enough!

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+
+**5.** Assuming you want to use postgres in your application, in your gemfile add the following line `gem 'pg'`. If you want a different database you just need to install the gem and define it on the `resources` section of `convox.yml`. More information about the databases we support [here](https://docs.convox.com/reference/primitives/app/resource#types)
```
beastawakens, line 70:
```suggestion
**5.** Assuming you want to use postgres in your application, in your gemfile add the following line `gem 'pg'`. If you want to use a different database you just need to install the gem and define it on the `resources` section of `convox.yml`. You can find information about the databases we currently natively support [here](https://docs.convox.com/reference/primitives/app/resource#types)
```

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+
+
+**6.** As mentioned in step **#2**, when you define a resource Convox will create an environment variable with that name for your application to access it. Given in our scenario the name of our resource is `database`, then it will generate a `DATABASE_URL`, which we will use to store/retrieve data. If you change the name of your resource, remember to also change the name of the variable to `[RESOURCE-name]_URL`. To use Convox' database resource as in our convox.yml, in your `config/database.yml` add the following line under `default`:
```
beastawakens, line 73: Maybe just link to our documentation on this? https://docs.convox.com/reference/primitives/app/resource#linking

docs/example-apps/rails.md
```diff
@@ -71,8 +112,14 @@ And you can deploy your application to production
 convox deploy
 ```
 
+You need to define the secret key for your production application
```
beastawakens, line 115: Should this happen before you deploy?

docs/example-apps/rails.md
```diff
@@ -1,45 +1,86 @@
...
+ 
+Your `database.yml` should look like [this](https://github.com/convox-examples/rails/blob/master/config/database.yml)
+
+### Running Locally
+
+Once you are all setup you can switch to your local rack with ```convox switch dev``` and from your project's folder you can start your local application with ```convox start```.
+
+You should now be able to access your application by going to [https://web.rails.convox](https://web.rails.convox). If you renamed anything you may need to modify your local URL. The format is https://[service name].[app name].convox
 
 ### Deploying to production
```
beastawakens, line 86: Probably need to add about switching to a Production Rack here...

beastawakens (changes requested):
Left a few comments and suggestions

docs/example-apps/rails.md
kaiomagalhaes: We have a 

```
And switch your CLI to your production rack

```bash
convox switch [rack name]
```
```

beastawakens (approved):

jasontclark (approved):

PR #351, opened 2021-07-06
aws: adding a timeout to subnet deletes.

jasontclark:
To prevent the `timeout while waiting for state to become 'destroyed'` error that occurs when deleting racks, I've added a 6-hr timeout value for the deletion of private and public subnets.

Signed-off-by: Jay Clark <jay@convox.com>

### DIFF
--- a/terraform/cluster/aws/network.tf
+++ b/terraform/cluster/aws/network.tf
@@ -41,6 +41,10 @@ resource "aws_subnet" "public" {
     "kubernetes.io/cluster/${var.name}" : "shared"
     "kubernetes.io/role/elb" : ""
   })
+
+  timeouts {
+    delete = "6h"
+  }
 }
 
 resource "aws_route_table" "public" {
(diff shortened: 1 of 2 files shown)


terraform/cluster/aws/network.tf
```diff
@@ -41,6 +41,10 @@ resource "aws_subnet" "public" {
     "kubernetes.io/cluster/${var.name}" : "shared"
     "kubernetes.io/role/elb" : ""
   })
+
+  timeouts {
```
kaiomagalhaes, line 45: @jasontclark shouldn't we add documentation about it somewhere?

kaiomagalhaes (changes requested):

terraform/cluster/aws/network.tf
jasontclark: @kaiomagalhaes whoops! yes, I'll add it.

kaiomagalhaes (approved):

PR #359, opened 2021-07-15
build: allow builds from outside the rack

ddollar:
adds `convox build --external` and `convox deploy --external` which
build the app images on the local machine and then push them up
into the rack's docker registry


beastawakens:
@ddollar is this ready to ship? Looks awesome!

ddollar:
I believe so. I've been using it to speed up my builds quite a lot by using buildx/buildkit external to the Rack.

PR #365, opened 2021-07-19
terraform: support for terraform 1.0

ddollar:
this change only addresses the aws and local providers


terraform/api/k8s/atom.tf
```diff
@@ -100,7 +100,7 @@ resource "kubernetes_deployment" "atom" {
           image_pull_policy = "Always"
 
           resources {
-            requests {
+            requests = {
```
beastawakens, line 103: @ddollar did you add these `=` in explicitly?  I don't think that's valid HCL for Terraform?

terraform/elasticsearch/k8s/main.tf
```diff
@@ -129,7 +129,7 @@ resource "kubernetes_stateful_set" "elasticsearch" {
           }
 
           resources {
-            requests {
+            requests = {
```
beastawakens, line 132: Same with this one

terraform/fluentd/k8s/main.tf
```diff
@@ -131,11 +131,11 @@ resource "kubernetes_daemonset" "fluentd" {
           }
 
           resources {
-            limits {
+            limits = {
```
beastawakens, line 134: Ditto

terraform/fluentd/k8s/main.tf
```diff
@@ -131,11 +131,11 @@ resource "kubernetes_daemonset" "fluentd" {
...
               memory = "200Mi"
             }
 
-            requests {
+            requests = {
```
beastawakens, line 138: Ditto

terraform/resolver/k8s/main.tf
```diff
@@ -155,7 +155,7 @@ resource "kubernetes_deployment" "resolver" {
           }
 
           resources {
-            requests {
+            requests = {
```
beastawakens, line 158: Same

terraform/router/nginx/main.tf
```diff
@@ -232,7 +232,7 @@ resource "kubernetes_deployment" "ingress-nginx" {
           }
 
           resources {
-            requests {
+            requests = {
```
beastawakens, line 235: Same

beastawakens (changes requested):
Not sure those blocks are valid anymore with the `=` in there...?

terraform/api/k8s/atom.tf
ddollar: Those = are required in the upgraded kubernetes provider

beastawakens: Aaah, okay, that would be why it failed with the non-upgraded provider 😉 😆 
Will have to manage this upgrade carefully for everyone on 0.13 provisioned racks.  Get them to 0.14 first and then up to 1.0

PR #373, opened 2021-07-23
docs: adding the workaround script to the Ubuntu local install docs.

jasontclark:


beastawakens:
Rather than have a meaty script clogging up the page, can we rather just host it somewhere (GitHub, S3, wherever) and direct the users to do a `source <(curl -s https://host/ubuntu-dns-patch.sh)` or similar?

Also, although it makes zero difference, let's call the script the ubuntu-dns-workaround or patch or something, we're not working around Convox, we're working around the bug in that library 😬

beastawakens:
Also, will it work for people without bash?  Could it just be a `#!/bin/sh` script?  Super edge case and if someone has replaced bash with fish or zsh or something they are probably advanced enough to figure it out but would be nice to not make them do that if there's no need.

kaiomagalhaes:
I second Ed's comment about creating a script and hosting it somewhere

Twsouza:
local racks is fixed now, see https://docs.convox.com/installation/development-rack/

PR #382, opened 2021-08-06
rack: terraform docker hub auth

beastawakens:


terraform/api/k8s/atom.tf
```diff
@@ -93,11 +93,13 @@ resource "kubernetes_deployment" "atom" {
...
+
         container {
-          name              = "system"
-          args              = ["atom"]
-          image             = "${var.image}:${var.release}"
-          image_pull_policy = "Always"
+          name               = "system"
+          args               = ["atom"]
+          image              = "${var.image}:${var.release}"
+          image_pull_policy  = "IfNotPresent"
```
jasontclark, line 102: Looks like this may have been this issue right here, is that true?

terraform/api/k8s/main.tf
```diff
@@ -105,7 +106,7 @@ resource "kubernetes_deployment" "api" {
           name              = "system"
           args              = ["api"]
           image             = "${var.image}:${var.release}"
-          image_pull_policy = "Always"
+          image_pull_policy = "IfNotPresent"
```
jasontclark, line 109: Same here, correct?

jasontclark (commented):
Took a look at this PR.  Looks like the `image_pull_policy` parameter was the main culprit.  This LGTM. 👍

kaiomagalhaes (approved):

PR #401, opened 2021-11-11
Configurable HA

heronrs:
Add the `high_availability` parameter and configure less resources(infra and k8s) replicas if it's disabled


terraform/cluster/aws/main.tf
```diff
@@ -105,7 +106,7 @@ resource "aws_eks_node_group" "cluster" {
   scaling_config {
     desired_size = 1
     min_size     = 1
-    max_size     = 100
+    max_size     = var.high_availability ? 100 : 3
```
heronrs, line 109: Leaving it at 3 when disabled as depending on the workloads it could run out of NICs for pods(t3 allows 12 per instance and rack + k8s/addon resources already get 13)

terraform/cluster/aws/main.tf
```diff
@@ -11,8 +11,9 @@ provider "kubernetes" {
 }
 
 locals {
-  oidc_sub           = "${replace(aws_iam_openid_connect_provider.cluster.url, "https://", "")}:sub"
-  availability_zones = var.availability_zones != "" ? compact(split(",", var.availability_zones)) : data.aws_availability_zones.available.names
+  availability_zones     = var.availability_zones != "" ? compact(split(",", var.availability_zones)) : data.aws_availability_zones.available.names
+  network_resource_count = var.high_availability ? 3 : 2
```
heronrs, line 15: EKS needs at least 2 subnets in different AZs

terraform/router/aws/main.tf
```diff
@@ -15,8 +15,10 @@ module "nginx" {
     kubernetes = kubernetes
   }
 
-  namespace = var.namespace
-  rack      = var.name
+  namespace    = var.namespace
+  rack         = var.name
+  replicas_max = var.high_availability ? 2 : 1
+  replicas_min = var.high_availability ? 10 : 1
```
beastawakens, line 21: Is this the right way round? With HA on, Max is 2 and Min is 10? 😄
What was the default behaviour before adding this?

beastawakens (changes requested):
This looks great! Only one code question.
Have you looked at how easy this would be to roll out to the other clouds?  Digital Ocean is a good candidate as it's the 'cheaper' cloud already.
Also, what's the experience like when you run this for real? (for Rack updates and app deployments)

ddoc (approved):

PR #402, opened 2021-11-15
Bump terraform deps to support arm64

nyergler:
Hashicorp packages need to be updated to support installing local racks
on arm64 machines.

### DIFF
--- a/terraform/api/k8s/versions.tf
+++ b/terraform/api/k8s/versions.tf
@@ -6,7 +6,7 @@ terraform {
     }
     random = {
       source  = "hashicorp/random"
-      version = "~> 2.2"
+      version = "~> 3.1"
     }
   }
   required_version = ">= 0.12"
--- a/terraform/cluster/azure/versions.tf
+++ b/terraform/cluster/azure/versions.tf
@@ -13,7 +13,7 @@ terraform {
     }
     random = {
       source  = "hashicorp/random"
-      version = "~> 2.2"
+      version = "~> 3.1"
     }
   }
   required_version = ">= 0.12"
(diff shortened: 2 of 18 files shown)


nyergler:
I'm still trying to figure out how to run the CLI locally with these changes; if you have pointers/suggestions, that'd be helpful!

heronrs:
Hi @nyergler , thanks for submitting the PR!
An easy way to test terraform changes is to first try installing a local rack `convox rack install local dev`, then press `ctrl + c` to interrupt the installation. After that go to the rack config directory(we use [xdg](https://github.com/adrg/xdg#xdg-base-directory) so the path for mac should be something like `~/Library/Application Support/convox/racks/your-rack-name`.

Open the `main.tf` file and replace the `source` URL with the commit/branch-name you pushed to your remote repo. For this specific PR it will look like `source = "github.com/polytomic/convox//terraform/system/local?ref=98ad9cd41d2b312e932a8b311c2b987607baab64"`.

Then just do a `terraform init && terraform apply`. 

After that, you can just play around with the `tf` files inside the rack config folder to test new changes.

nyergler:
Thanks, @heronrs .

We did a little digging and found the Terraform files you mentioned; also discovered that the following would work to bootstrap the local rack --

```
$ CONVOX_TERRAFORM_SOURCE="github.com/polytomic/convox//terraform/system/%s?ref=arm64-terraform-deps" convox rack install local dev
```

We've been able to configure and run a local rack this way, and haven't encountered any Terraform-related issues yet.

heronrs:
Awesome, glad to hear it!
I'll also do some tests in the following days and proceed with the merge.
Thanks again for submitting this PR.

heronrs:
@nyergler I also bumped the `null` and `local` providers as we detected it was also needed when installing cloud racks so I created a new PR to be able to add more changes. ref #403

heronrs:
https://github.com/convox/convox/releases/tag/3.1.1

PR #407, opened 2021-12-08
Fix rack instalation on GCP

heronrs:
Removes deprecated basic auth from GKE cluster and bump the google provider so it works on arm64 - this last one needed some additional changes to resource parameters

### DIFF
--- a/pkg/rack/terraform.go
+++ b/pkg/rack/terraform.go
@@ -160,7 +160,7 @@ func (t Terraform) Endpoint() (*url.URL, error) {
 	return url.Parse(t.endpoint)
 }
 
-func (t Terraform) Latest() (string, error) {
+func (Terraform) Latest() (string, error) {
 	return terraformLatestVersion()
 }
 
@@ -222,7 +222,7 @@ func (t Terraform) Provider() string {
 	return t.provider
 }
 
-func (t Terraform) Remote() bool {
+func (Terraform) Remote() bool {
 	return false
 }
 
@@ -231,21 +231,16 @@ func (t Terraform) Status() string {
 }
 
 func (t Terraform) Uninstall() error {
-	env, err := terraformEnv(t.provider)
-	if err != nil {
-		return err
-	}
-
 	dir, err := t.ctx.SettingDirectory(fmt.Sprintf("racks/%s", t.name))
 	if err != nil {
 		return err
 	}
 
-	if err := terraform(t.ctx, dir, env, "init", "-no-color", "-upgrade"); err != nil {
+	if err := terraform(t.ctx, dir, "init", "-no-color", "-upgrade"); err != nil {
 		return err
 	}
 
-	if err := terraform(t.ctx, dir, env, "destroy", "-auto-approve", "-no-color", "-refresh=true"); err != nil {
+	if err := terraform(t.ctx, dir, "destroy", "-auto-approve", "-no-color", "-refresh=true"); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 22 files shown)


terraform/api/gcp/storage.tf
```diff
@@ -7,6 +7,8 @@ resource "random_string" "suffix" {
 resource "google_storage_bucket" "storage" {
   name = "${var.name}-storage-${random_string.suffix.result}"
 
-  bucket_policy_only = true
```
heronrs, line 10: deprecated and replaced by `uniform_bucket_level_access`

terraform/cluster/gcp/main.tf
```diff
@@ -22,24 +20,19 @@ resource "google_container_cluster" "rack" {
   }
 
   workload_identity_config {
-    identity_namespace = "${data.google_project.current.project_id}.svc.id.goog"
```
heronrs, line 25: deprecated and replaced by `workload_pool`

terraform/cluster/gcp/main.tf
```diff
@@ -59,7 +52,7 @@ resource "google_container_node_pool" "rack" {
     }
 
     workload_metadata_config {
-      node_metadata = "GKE_METADATA_SERVER"
```
heronrs, line 62: deprecated and replaced by `mode`

pkg/rack/terraform.go
```diff
@@ -594,7 +594,13 @@ func terraformEnv(provider string) (map[string]string, error) {
 	case "do":
 		return requireEnv("DIGITALOCEAN_ACCESS_ID", "DIGITALOCEAN_SECRET_KEY", "DIGITALOCEAN_TOKEN")
 	case "gcp":
-		return requireEnv("GOOGLE_CREDENTIALS", "GOOGLE_PROJECT")
+		{
+			gap := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
heronrs, line 598: just to keep it backward compatible

pkg/rack/terraform.go
```diff
@@ -728,7 +728,11 @@ func terraformWriteTemplate(filename, version string, params map[string]interfac
 		}
 
 		output "provider" {
-			value = "{{.Provider}}"
+			value ="{{.Provider}}"
+		}
+
+		output "release" {
```
heronrs, line 734: `console` will use this when uninstalling a rack. currently, there's a bug because this output variable is not set so it will always try to uninstall using the latest changes in `master` which can have unexpected consequences.

pkg/rack/terraform.go
```diff
@@ -245,7 +245,7 @@ func (t Terraform) Uninstall() error {
 		return err
 	}
 
-	if err := terraform(t.ctx, dir, env, "destroy", "-auto-approve", "-no-color"); err != nil {
+	if err := terraform(t.ctx, dir, env, "destroy", "-auto-approve", "-no-color", "-refresh=true"); err != nil {
```
heronrs, line 248: Ensure that we always refresh the state before destroying the rack

terraform/system/gcp/variables.tf
```diff
@@ -15,7 +15,7 @@ variable "name" {
 }
 
 variable "node_type" {
-  default = "n1-standard-1"
+  default = "n1-standard-2"
```
heronrs, line 18: `n1-standard-1` has only 1CPU and the `fluentd daemonset` will fail to be scheduled on the `elasticsearch` node for lack of CPU resources.

pkg/rack/terraform.go
```diff
@@ -557,7 +547,7 @@ func requireEnv(vars ...string) (map[string]string, error) {
 	return env, nil
 }
 
-func terraform(c *stdcli.Context, dir string, env map[string]string, args ...string) error {
```
heronrs, line 560: `env` is not used

beastawakens (approved):
Let's ensure we have the Console update rolled out before we release this, and communicate with self-hosted Console users to update as well.

PR #408, opened 2021-12-14
Validate schedule rack down and up parameters on update

Twsouza:
To set the ScheduleRackDown and ScheduleRackUp, the user needs to pass them both, if not it'll return an error.

image

### DIFF
--- a/pkg/cli/rack.go
+++ b/pkg/cli/rack.go
@@ -88,6 +88,19 @@ func init() {
 	})
 }
 
+func validateParams(params map[string]string) error {
+	if params["high_availability"] != "" {
+		return errors.New("the high_availability parameter is only supported during rack installation")
+	}
+
+	srdown, srup := params["ScheduleRackScaleDown"], params["ScheduleRackScaleUp"]
+	if (srdown == "" || srup == "") && (srdown != "" || srup != "") {
+		return errors.New("to schedule your rack to turn on/off you need both ScheduleRackScaleDown and ScheduleRackScaleUp parameters")
+	}
+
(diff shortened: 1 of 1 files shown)


heronrs (approved):
LGTM

pkg/cli/rack.go
```diff
@@ -88,6 +88,19 @@ func init() {
 	})
 }
 
+func validateParams(params map[string]string) error {
+	if params["high_availability"] != "" {
+		return errors.New("the high_availability parameter is only supported during rack installation")
+	}
+
+	srdown, srup := params["ScheduleRackScaleDown"], params["ScheduleRackScaleUp"]
```
beastawakens, line 96: For v3, the Rack params should use Snake Case rather than Camel Case 
```suggestion
	srdown, srup := params["schedule_rack_scale_down"], params["schedule_rack_scale_up"]
```

beastawakens (changes requested):

pkg/cli/rack.go
Twsouza: @beastawakens The ScheduleRackScaleDown and ScheduleRackScaleUp are only going to be used by v2 rack.
This validation is in case someone with v2 rack tries to update through the cli v3.

beastawakens: Okay, fair enough.  We should also look to add the other v2 params with similar restrictions (`VPCCIDR` and `Private` spring to mind) at a future point, but no need to complicate this PR with it for now.

beastawakens (approved):

PR #409, opened 2021-12-14
rack: Add arbitrary security context to allow for token reads

beastawakens:
This will need a little bit of testing, but should solve https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8


heronrs (approved):
LGTM. Does `65534` has any special meaning?

Twsouza:
@beastawakens @heronrs since this problem will be solved in the EKS version 1.19 that we will release very soon (https://github.com/convox/convox/pull/411), should we close this PR?

beastawakens:
Probably not needed anymore, unless someone is insisting on staying on <3.3.x (i.e. k8s 1.18 or below)...

PR #411, opened 2022-01-09
K8s update 1.19

heronrs:
Updates k8s versions to 1.19 on AWS and Azure.
GCP and DO are not necessary since they will always use the latest available release.

Also dropped `xgo` for cross-compilation and build binaries without CGO. This will improve portability since the binary will be statically linked(e.g will not depend on underlying OS libraries(for almost all cases).

Depends on https://github.com/convox/convox/pull/338


Twsouza (dismissed):
lgtm

Twsouza (approved):
:+1:

PR #416, opened 2022-01-21
Add AWS spot  instances support

heronrs:
Adds the necessary parameters to enable the creation of racks with spot instances.
I also changed `node_type` to accept a comma-separated list of instance types since AWS recommends specifying multiple instance types when using spot instances

### DIFF
--- a/terraform/system/aws/variables.tf
+++ b/terraform/system/aws/variables.tf
@@ -42,7 +42,7 @@ variable "name" {
 }
 
 variable "node_capacity_type" {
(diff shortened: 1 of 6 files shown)


terraform/system/aws/variables.tf
```diff
@@ -41,6 +41,10 @@ variable "name" {
   type = string
 }
 
+variable "node_capacity_type" {
+  default = "on_demand"
```
heronrs, line 45: lower case just to keep consistency with the other variables

PR #417, opened 2022-01-27
Add rack parameter to install it on an existing vpc

heronrs:
Adds new rack install parameters `vpc_id`  and `internet_gateway_id` if users want to use an existing VPC to host the rack.


Twsouza (approved):
:+1:

beastawakens (approved):
🚢

PR #418, opened 2022-02-02
k8s: add ability to specify annotations for custom load balancers

beastawakens:


provider/k8s/template/app/balancer.yml.tmpl
```diff
@@ -6,6 +6,9 @@ metadata:
   annotations:
     service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
     service.beta.kubernetes.io/aws-load-balancer-type: nlb
+    {{ range keyValue .Annotations }}
```
heronrs, line 9: Didn't test this but shouldn't it be something like
```
    {{ range $key, $value := .Annotations }}
    {{.key}}: "{{ quoteEscape .value}}"
    {{ end }}
```

Maybe you're just using a syntax I'm not aware though

beastawakens: Yeah, `keyValue` is a helper function from: https://github.com/convox/convox/blob/master/provider/k8s/template.go#L50-L67
to return a range of objects containing the `Key` and `Value`.  Makes it easier to work with in templates like this 😊

heronrs (approved):
LGTM

PR #421, opened 2022-02-11
Build the binary as statically linked

Twsouza:
The binary was not being built as statically linked but as dynamically linked.
This was causing errors on Alpine.

### DIFF
--- a/cmd/convox/Makefile
+++ b/cmd/convox/Makefile
@@ -7,12 +7,10 @@ all: build
 
 build:
 	mkdir pkg/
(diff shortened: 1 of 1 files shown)


heronrs (approved):
Great Find!

mtcmorris:
FYI I found on my Mac M1 (using the darwin/arm64 binary) that `CGO_ENABLED=0` caused https://github.com/convox/go-u2fhost to not detect my usb yubikey.  Perhaps just on my machine but I'd always get the error `no valid tokens found`

PR #424, opened 2022-03-17
Add local Kubernetes version to Known Issues

niclas-ahden:
Docker Desktop defaults to the latest Kubernetes version which prevents new local racks from being created. Mentioning this here, and it should probably also be cross-referenced in https://github.com/convox/convox/blob/master/docs/installation/development-rack/macos.md.


heronrs:
Thanks Niclas! I also added the cross-reference link on the macOS setup page.

PR #426, opened 2022-03-29
Test and fix for proxy dropping packets

slaskis:
The broken proxying in a v3 rack is causing a lot of frustrations for me but hopefully this failing test might help you figure out what the cause is. 

convox/issues#9

### DIFF
--- a/pkg/api/proxy_test.go
+++ b/pkg/api/proxy_test.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"strings"
+	"math/rand"
 	"testing"
 
 	"github.com/convox/convox/pkg/structs"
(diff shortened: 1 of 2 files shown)


slaskis:
Weirdly enough making this buffer smaller (I've tested 512 and 256) seems to make this test pass:

https://github.com/convox/convox/blob/master/vendor/github.com/convox/stdsdk/client.go#L236

slaskis:
I believe I found it! 

After hours of debugging and adding logs everywhere I've distilled it to this:

The issue was that a TextMessage io.Reader might not be fully read for whatever reason, so the rest of the data will be dropped with the implementation in master. This PR will use `io.CopyBuffer` to make sure it reads until EOF.

heronrs:
@slaskis can you confirm if the following is working for you?
I basically built a new image from this branch and replaced on the API pod. trying to exec into processes just hangs at the prompt:
```
c exec web-77db575f89-bnpxl bash -a console
console@web-77db575f89-bnpxl:/$
```
and it just won't respond anymore.

slaskis:
Ah interesting. I have only tried it on the client side (using a build from this PR when I want to proxy successfully) and didn't even realize that it's also used in an api pod...

is that why the build/convox/test checks are not done too you think?

nightfury1204:
old pr

PR #427, opened 2022-03-29
Add information about docker hub registry and troubleshooting on macOS

Twsouza:
When adding DockerHub to the registries, users must use `https://`. See [Pull an Image from a Private Registry](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-by-providing-credentials-on-the-command-line).

Documentation about the troubleshooting for local rack DNS issues on macOS.


heronrs (approved):
lgtm

PR #428, opened 2022-04-01
cli: Correctly parse GH releases response

heronrs:
Correctly parse the GH releases response body when `convox rack update`


Twsouza (approved):
:+1:

PR #432, opened 2022-04-11
Add sh to code tag

Twsouza:
Add the `sh` to the code block to style it correctly. Right now the code is incomprehensible:

image

Applying the style:

image


heronrs (approved):
:fire:

PR #433, opened 2022-04-12
When canceling an app update, rewrite the last active release on top

heronrs:
When canceling an app update, rewrite the last active release on top of the canceled one.

### DIFF
--- a/pkg/cli/apps.go
+++ b/pkg/cli/apps.go
@@ -109,12 +109,32 @@ func Apps(rack sdk.Interface, c *stdcli.Context) error {
 }
 
 func AppsCancel(rack sdk.Interface, c *stdcli.Context) error {
-	app := coalesce(c.Arg(0), app(c))
+	aname := coalesce(c.Arg(0), app(c))
 
-	c.Startf("Cancelling deployment of <app>%s</app>", app)
+	c.Startf("Cancelling deployment of <app>%s</app>", aname)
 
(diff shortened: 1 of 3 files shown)


Twsouza (commented):
@heronrs should we add a description to the canceled release, something like `Release canceled - ...`? Or some way that we can identify that the build was canceled?

pkg/cli/apps.go
```diff
@@ -109,13 +109,33 @@ func Apps(rack sdk.Interface, c *stdcli.Context) error {
...
+	if err != nil {
 		return err
 	}
+	for _, r := range rs {
+		if app.Release == r.Id {
+			_, err := rack.ReleaseCreate(app.Name, structs.ReleaseCreateOptions{Build: &r.Build, Description: &r.Description, Env: &r.Env})
+			if err != nil {
+				return err
+			}
+		}
```
Twsouza, line 137: maybe add a break statement when finding the release.

Twsouza (changes requested):

heronrs:
> @heronrs should we add a description to the canceled release, something like `Release canceled - ...`? Or some way that we can identify that the build was canceled?

@Twsouza I think this one would be nice to have but since it involves a bit more work - we'd need to create a rack endpoint specifically for updating a release - I think we can do it later in a different task.

Twsouza (approved):
:fire:

PR #440, opened 2022-05-23
AWS - Add proxy protocol parameter (client's real source ip)

heronrs:
Adds a new parameter for AWS racks to enable [proxy protocol ](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol).
With the protocol enabled, convox apps will be able to get the client source IP using the `x-forwarded-for` header key.

The parameter will toggle proxy protocol on/off in the NLB and Nginx Deployment.
Changing the parameter will require a 5 - 10 minutes Downtime which is the time needed for the NLB to start sending the proxy protocol header and Nginx restarting with the new config and be able to parse the header.

### DIFF
--- a/docs/installation/production-rack/aws.md
+++ b/docs/installation/production-rack/aws.md
@@ -54,6 +54,7 @@ The following environment variables are required:
 | **node_disk**            | **20**                 | Node disk size in GB                                                                                           |
 | **node_type**            | **t3.small**           | Node instance type. You can also pass a comma separated list of instance types                                 |
 | **private**              | **true**               | Put nodes in private subnets behind NAT gateways                                                               |
+| **proxy_protocol**       | **true**               | Enable proxy protocol. With this parameter set, the client source ip will be available in the request header `x-forwarded-for` key. **Requires 5 - 10 minutes downtime**.          |
 | **region**               | **us-east-1**          | AWS Region                                                                                                     |
 | **syslog**               |                        | Forward logs to a syslog endpoint (e.g. **tcp+tls://example.org:1234**)                                        |
 | **vpc_id** *             |                        | Use an existing VPC for cluster creation. Make sure to also pass the **cidr** block and **internet_gateway_id**|
--- a/terraform/rack/aws/variables.tf
+++ b/terraform/rack/aws/variables.tf
@@ -34,6 +34,10 @@ variable "oidc_sub" {
   type = string
 }
 
+variable "proxy_protocol" {
+  default = false
+}
+
 variable "release" {
   type = string
 }
--- a/terraform/router/aws/proxy-protocol.sh
+++ b/terraform/router/aws/proxy-protocol.sh
@@ -2,10 +2,36 @@
 
 set -e
 
+if [ ! -x "$(which jq)" ]
+then
+    echo "missing jq binary"
+    exit 1
+fi
+
+if [ ! -x "$(which aws)" ]
+then
+    echo "missing aws cli"
+    exit 1
+fi
+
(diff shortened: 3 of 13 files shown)


terraform/router/nginx/main.tf
```diff
@@ -167,12 +167,13 @@ resource "kubernetes_deployment" "ingress-nginx" {
...
-          system  = "convox"
-          service = "ingress-nginx"
-          type    = "service"
+          app            = "system"
+          name           = "ingress-nginx"
+          rack           = var.rack
+          system         = "convox"
+          service        = "ingress-nginx"
+          type           = "service"
+          proxy_protocol = var.proxy_protocol
```
heronrs, line 176: this new label will force the Deployment to restart to pick up the Nginx config-map update.

terraform/router/aws/proxy-protocol.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
heronrs, line 1: This script enables proxy protocol in the NLB. We can't use service annotations to enable it because the in-tree AWS Load Balancer implementation doesn't support it.

Twsouza (dismissed):
Good catch! 🔥 
We have to start planning the migration of in-tree to out-tree. [Removing In-Tree Cloud Provider Code](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cloud-provider/2395-removing-in-tree-cloud-providers#phase-4---disabling-in-tree-providers)

> For GA, the feature gate will be enabled by default and locked. Users at this point MUST migrate to external components and use of the in-tree cloud providers will be disallowed. One release after GA, the in-tree cloud providers can be safely removed. GA is targeted for v1.25 or v1.26.

terraform/router/nginx/main.tf
```diff
@@ -1,12 +1,13 @@
 resource "kubernetes_config_map" "nginx-configuration" {
+  count = var.cloud_provider == "aws" ? 0 : 1
```
Twsouza, line 2: what is this variable used for?

heronrs: 0 means the resource won't be created and 1 it will be created.
In the case of aws we are now managing the configmap on the aws module.

Twsouza (approved):
🚢

terraform/rack/aws/variables.tf
```diff
@@ -34,6 +34,10 @@ variable "oidc_sub" {
   type = string
 }
 
+variable "proxy_protocol" {
+  default = false
```
beastawakens, line 38: @heronrs should this be true?  The docs suggest that the default is `true`.  Or should the docs change?

docs/installation/production-rack/aws.md
```diff
@@ -54,6 +54,7 @@ The following environment variables are required:
 | **node_disk**            | **20**                 | Node disk size in GB                                                                                           |
 | **node_type**            | **t3.small**           | Node instance type. You can also pass a comma separated list of instance types                                 |
 | **private**              | **true**               | Put nodes in private subnets behind NAT gateways                                                               |
+| **proxy_protocol**       | **true**               | Enable proxy protocol. With this parameter set, the client source ip will be available in the request header `x-forwarded-for` key. **Requires 5 - 10 minutes downtime**.          |
```
beastawakens, line 57: Should this be `false`?

heronrs: Yep, correct, default should be false to keep backward compatibility. Thanks for spotting that

PR #442, opened 2022-05-26
proxy-protocol script: Consider aws region and improve timeout logic

heronrs:
Improve logic for waiting NLB provisioning and also pass the select region to aws cli.

### DIFF
--- a/terraform/router/aws/proxy-protocol.sh
+++ b/terraform/router/aws/proxy-protocol.sh
@@ -20,18 +20,24 @@ then
     exit 1
 fi
 
-sleep 5 # make sure the load-balancer and target groups associated with the router service were already created 
-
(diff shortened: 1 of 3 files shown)


terraform/router/aws/proxy-protocol.sh
```diff
@@ -20,18 +20,24 @@ then
...
+while [ -z "$target_groups" ]
+do
+    seconds=$(($seconds+30))
+    if [ $seconds -ge 600 ]
+    then
+        echo "failed to provision load-balancer for $1"
+        exit 1
+    fi
+    echo "waiting for NLB target-groups..."
+    sleep 30
```
ddoc, line 36: can you do a loop here waiting for the target groups to be available?

Twsouza (approved):
🔥

PR #444, opened 2022-06-06
Adding more details about AutoScaling

Twsouza:
k8s use the `1000m` as 1 vCPU instead of 1024. See [CPU resource units](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu)

> Note: Kubernetes doesn't allow you to specify CPU resources with a precision finer than 1m. Because of this, it's useful to specify CPU units less than 1.0 or 1000m using the milliCPU form; for example, 5m rather than 0.005.

And also added more details about how the HPA works, using the [official documentation](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/)

### DIFF
--- a/docs/deployment/scaling.md
+++ b/docs/deployment/scaling.md
@@ -9,9 +9,9 @@ url: /deployment/scaling
 Convox allows you to easily scale any [Service](/reference/primitives/app/service) on the following dimensions:
 
 - Horizontal concurrency (number of [Processes](/reference/primitives/app/process))
-- CPU allocation (in CPU units where 1024 units is one full CPU)
+- CPU allocation (in CPU units where 1000 units is one full CPU)
 - Memory allocation (in MB)
- 
+
 ## Initial Defaults
 
 You can specify the scale for any [Service](/reference/primitives/app/service) in your [convox.yml](/configuration/convox-yml)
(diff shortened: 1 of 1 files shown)


docs/deployment/scaling.md
```diff
@@ -60,5 +60,10 @@ target values for CPU and Memory utilization (in percent):
...
 ```
-The number of [Processes](/reference/primitives/app/process) will be continually adjusted to maintain
-your target metrics.
\ No newline at end of file
+The number of [Processes](/reference/primitives/app/process) will be continually adjusted to maintain your target metrics.
+
+You must consider that the targets for CPU and Memory use the pods' limits to calculate the utilization percentage. So if you set the target for CPU as `70` and have two pods, it will trigger the HPA only if the utilization percentage sum divided by the pods' count is bigger than 70%. The desired replicas will be calculated to satisfy the percentage. Being the `currentMetricValue` computed by taking the average of the given metric across all Pods in the HorizontalPodAutoscaler's scale target.
+
+```html
+desiredReplicas = ceil[currentReplicas * ( currentMetricValue / desiredMetricValue )]
```
heronrs, line 68: Can we try rewriting this without using Kubernetes primitives?

>You must consider that the targets for CPU and Memory use the service replicas limit to calculate the utilization percentage. So if you set the target for CPU as `70` and have two replicas, it will trigger the Auto Scale only if the utilization percentage sum divided by the replica's count is bigger than 70%. The desired replicas will be calculated to satisfy the percentage. The current target value is computed by taking the average of the given metric across all service replicas

heronrs (approved):

PR #445, opened 2022-06-09
Add new rack sync cli command to force syncing the v2 racks API url

heronrs:
New cli command `convox rack sync` to prepare for the v2 Classic x ALB migration.

This command allow users to manually sync the rack API URL. Ref https://github.com/convox/console/pull/222


Twsouza (dismissed):
🔥

PR #446, opened 2022-06-14
Prevent skip minor on update

Twsouza:
It will receive an error if tries to skip one minor version (e.g. 3.5.X to 3.7.X) and will ask to update to the next minor release (e.g. 3.6.X) first.


heronrs (approved):
LGTM

PR #447, opened 2022-06-17
feature/344: Fetch nodes and processes cpu and mem usage using metrics api

nightfury1204:
This PR provides backend support for : https://github.com/convox/issues-private/issues/344

It fetches the nodes and processes memory and cpu usage using metrics api. mertic-server installed in k8s cluster is responsible for that.


provider/k8s/helpers.go
```diff
@@ -314,3 +315,27 @@ func volumeTo(v string) (string, error) {
...
+	// 1000m (milicores) = 1 core = 1 vCPU = 1 AWS vCPU = 1 GCP Core
+	return float64(millicore) / 1000
+}
+
+func toMemMB(bytes int64) float64 {
+	// 1024*1024 bytes = 1 MiB
+	return float64(bytes) / (1024 * 1024)
+}
+
+func calculatePodCpuAndMem(m *metricsv1beta1.PodMetrics) (cpu float64, mem float64) {
```
Twsouza, line 329: This function is not meant to be a helper, it's a provider function, perhaps it's better to move it to be a Provider function, something like `p.CalculatePodCpuAndMem`. Since you have the metrics interface in the Provider struct, you won't need to pass it to the function.

Twsouza (changes requested):
It's only returning the sum of the pods' metrics. It would be good to have another way to get the nodes' metrics (even in a different endpoint).

go.mod
```diff
@@ -55,4 +55,5 @@ require (
 	k8s.io/client-go v0.16.6
 	k8s.io/code-generator v0.16.6
 	k8s.io/klog v1.0.0
+	k8s.io/metrics v0.16.6
```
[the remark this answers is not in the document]
nightfury1204: this version is tight to the client-go version which is `v0.16.6`, to update it to latest version we need to update all the following related library too:
```
        k8s.io/api v0.16.6
	k8s.io/apimachinery v0.16.6
	k8s.io/client-go v0.16.6
	k8s.io/code-generator v0.16.6
```
Also codes needs to be updated too to use these updated packages, not sure how much though. That's why i didn't use the latest version.
I think it should be done in a seperate PR, because it could be a major change.

nightfury1204:
> It's only returning the sum of the pods' metrics. It would be good to have another way to get the nodes' metrics (even in a different endpoint).

@Twsouza it actually doesn't return the sum of the pods' metrics, it's returning sum of container metrics for each pod, because a pod can have multiple containers. This pr doesn't provide implementation for nodes metrics.

provider/k8s/helpers.go
nightfury1204: `m *metricsv1beta1.PodMetrics` this parameter is struct, it contains containers usage metrics. it differs pod to pod, so i have to pass it as parameter. It basically sums the resource usage and return it. That's why i put it in the helpers method.
But it can be made provider function too, but the parameter still will be there.
Should i make it provider function, what's your call?

Twsouza: No need, leave as it is 👍

Twsouza (approved):
👍

provider/k8s/process.go
```diff
@@ -150,6 +159,21 @@ func (p *Provider) ProcessList(app string, opts structs.ProcessListOptions) (str
 		pss = append(pss, *ps)
 	}
 
+	// ignore error because this api only exists if metrics crds are installed in the k8s
+	ms, err := p.MetricsClient.MetricsV1beta1().PodMetricses(p.AppNamespace(app)).List(am.ListOptions{LabelSelector: strings.Join(filters, ",")})
+	if err == nil {
```
heronrs, line 164: I think we can return the error here. It could be useful to notify the client about problems. We include metrics server by default in the rack instalation so if it's missing we'd know.

provider/k8s/process.go
```diff
@@ -150,6 +159,21 @@ func (p *Provider) ProcessList(app string, opts structs.ProcessListOptions) (str
...
+		metricsByPod := map[string]metricsv1beta1.PodMetrics{}
+		for _, m := range ms.Items {
+			metricsByPod[m.Name] = m
+		}
+
+		for i := range pss {
+			if m, has := metricsByPod[pss[i].Id]; has && len(m.Containers) > 0 {
```
heronrs, line 171: just checking. process.Id here is equal the pod name, right?

Twsouza:
@nightfury1204 is this only applied to processes? how is the instance going to be applied?

heronrs:
> @nightfury1204 is this only applied to processes? how is the instance going to be applied?

@nightfury1204 I think we'll need a similar endpoint for getting the metrics of the instance nodes as it's one of the tabs in the console UI.

It's fine if you want to do it in a different PR.

provider/k8s/process.go
nightfury1204: yes, in the existing code I saw using pod name used as id. ref: https://github.com/convox/convox/blob/master/provider/k8s/process.go#L591

nightfury1204:
> > @nightfury1204 is this only applied to processes? how is the instance going to be applied?
> 
> @nightfury1204 I think we'll need a similar endpoint for getting the metrics of the instance nodes as it's one of the tabs in the console UI.
> 
> It's fine if you want to do it in a different PR.

Got it. In that case, I will add the node usages metrics in this PR. That way we all be in the same context and they will get merged together.
I am putting label **do not merge** label for now. Once I pushed the changes I will remove it.

cc: @heronrs @Twsouza

provider/k8s/helpers.go
```diff
@@ -314,3 +315,27 @@ func volumeTo(v string) (string, error) {
...
+	if m == nil {
+		return 0, 0
+	}
+
+	cpuTotal := int64(0)
+	memTotal := int64(0)
```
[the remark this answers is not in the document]
nightfury1204: addressed it in next commit

provider/k8s/process.go
nightfury1204: addressed it in the next commit

pkg/cli/instances.go
```diff
@@ -42,7 +42,13 @@ func Instances(rack sdk.Interface, c *stdcli.Context) error {
 	t := c.Table("ID", "STATUS", "STARTED", "PS", "CPU", "MEM", "PUBLIC", "PRIVATE")
 
 	for _, i := range is {
-		t.AddRow(i.Id, i.Status, common.Ago(i.Started), fmt.Sprintf("%d", i.Processes), common.Percent(i.Cpu), common.Percent(i.Memory), i.PublicIp, i.PrivateIp)
+		cpuPercent, memPercent := common.Percent(i.Cpu), common.Percent(i.Memory)
+		// for rack version 3, if allocatable metrics available
```
heronrs, line 46: #nit you can remove this comment as this repo only contains the code for v3 racks.

heronrs (approved):
LGTM. minor nit pick

pkg/cli/instances.go
nightfury1204: removed it

heronrs (approved):
lgtm :fire:

Twsouza (approved):
🚢

PR #448, opened 2022-06-27
Fix gcp rack instalation. Pin k8s version

heronrs:
After fixing the subnet quota problem I noticed that the GCP installation is still failing because the minimum version on the `REGULAR` release channel is 1.22 - which deprecated some API objects that we still use.

I changed the version to be statically managed so we guarantee it won't be automatically updated to 1.22


Twsouza (approved):
🔥

PR #452, opened 2022-07-12
Changing message when user tries to update directly

Twsouza:
This error happens when the console doesn't find the state file.

Also can happen if tries to update through RACK_URL.

### DIFF
--- a/provider/k8s/system.go
+++ b/provider/k8s/system.go
@@ -108,9 +108,9 @@ func (p *Provider) SystemReleases() (structs.Releases, error) {
 }
 
 func (p *Provider) SystemUninstall(name string, w io.Writer, opts structs.SystemUninstallOptions) error {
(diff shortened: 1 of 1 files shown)


nightfury1204:
This message change is for developer or user?
I am not sure user will understand the part `don't use RACK_URL or try moving your rack locally to uninstall`

Twsouza:
> This message change is for developer or user? I am not sure user will understand the part `don't use RACK_URL or try moving your rack locally to uninstall`

@nightfury1204 The `RACK_URL` is not documented anywhere, would be more for developers.

The `try moving your rack locally to uninstall` would be more useful to the user. They should follow this documentation https://docs.convox.com/management/console-rack-management/#moving-your-rack-from-the-console.

nightfury1204 (approved):

PR #453, opened 2022-07-15
Use launch template to use gp3

Twsouza:
Using a launch template, the EKS cluster nodes will use the gp3.

[Still testing]


heronrs (approved):
nice!

nightfury1204 (approved):
:fire:

PR #455, opened 2022-07-22
Fix for CLI rack update command

Twsouza:
Using a non-standard 3.X.X version breaks the CLI. It will try to split the numbers but will return an empty slice. Trying to access index 1 breaks the CLI.

Example: with a CLI on 3.5.3 and try to update to [gp3-default](https://github.com/convox/convox/releases/tag/gp3-default).

### DIFF
--- a/pkg/rack/console.go
+++ b/pkg/rack/console.go
@@ -256,11 +256,20 @@ func (c Console) UpdateVersion(version string) error {
 }
 
 func isSkippingMinor(currentVersion, version string) error {
-	cmv, err := strconv.Atoi(strings.Split(currentVersion, ".")[1])
+	splittedCurrent := strings.Split(currentVersion, ".")
+	if len(splittedCurrent) <= 1 {
+		return nil
+	}
+	cmv, err := strconv.Atoi(splittedCurrent[1])
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/rack/console.go
```diff
@@ -256,7 +256,12 @@ func (c Console) UpdateVersion(version string) error {
 }
 
 func isSkippingMinor(currentVersion, version string) error {
-	cmv, err := strconv.Atoi(strings.Split(currentVersion, ".")[1])
+	splittedVersion := strings.Split(currentVersion, ".")
+	if len(splittedVersion) == 0 {
+		return nil
+	}
+
+	cmv, err := strconv.Atoi(splittedVersion[0])
```
nightfury1204, line 264: shouldn't it be `splittedVersion[1]` instead of `splittedVersion[0]`

nightfury1204: also i think we need slice len check in here too `mv, err := strconv.Atoi(strings.Split(version, ".")[1])`

Twsouza: thanks for the feedback, I forgot about it 😄

nightfury1204 (approved):

heronrs (approved):

PR #457, opened 2022-07-27
Update deprecated k8s resources

heronrs:


provider/k8s/k8s.go
```diff
@@ -258,9 +273,32 @@ func (p *Provider) initializeTemplates() error {
...
+			time.Sleep(time.Second * 20)
+
+			// println(">>> re-applying cert-manager resources")
+			err := p.applySystemTemplate("cert-manager", nil)
+			if err != nil {
+				return err
+			}
+
+			println("update finished")
+		} else if d == nil {
```
heronrs, line 296: If we don't find the `cert-manager` deployment it means it's a new install.

provider/k8s/k8s.go
```diff
@@ -258,9 +273,32 @@ func (p *Provider) initializeTemplates() error {
...
 	}
 
+	d, _ := p.Cluster.AppsV1().Deployments("cert-manager").Get("cert-manager", am.GetOptions{})
+
+	// println(d.Spec.Template.Labels["app.kubernetes.io/version"])
+
 	if p.CertManager {
-		if err := p.applySystemTemplate("cert-manager", nil); err != nil {
-			return errors.WithStack(err)
+		if d != nil && d.Spec.Template.Labels["app.kubernetes.io/version"] != "v1.9.0" {
```
heronrs, line 281: if we re-apply the manifest on top of existing resources, the operation will hang forever so only apply it if running a different cert-manager version

provider/k8s/k8s.go
```diff
@@ -258,9 +273,32 @@ func (p *Provider) initializeTemplates() error {
...
+			println("updating cert-manager...")
+			p.deleteSystemTemplate("cert-manager", nil)
```
heronrs, line 283: i'm not checking for errors because there's a race condition between the deletion of the namespace and the deletion of the resources in the namespace.i.e delete command might return a "not found" error when deleting some resources bc it was already cascade deleted due to deleting the namespace.

provider/k8s/k8s.go
```diff
@@ -258,9 +273,32 @@ func (p *Provider) initializeTemplates() error {
...
+
+			// println(">>> waiting for namespace to be deleted")
+			// TODO Query the cert-manager namespace and return when it's deleted
+			time.Sleep(time.Second * 20)
```
heronrs, line 287: we should poll for the namespace and proceed when it's not found, i.e deleted.
`n, err := p.Cluster.CoreV1().Namespaces().Get("cert-manager", am.GetOptions{})`

`n == nil` means the namespace was not found and we can proceed.

PR #458, opened 2022-07-29
Fix environment variable bug when using convox apps cancel

heronrs:
Same logic as https://github.com/convox/rack/pull/3556


Twsouza (dismissed):
🚀

Twsouza (dismissed):
👍

PR #463, opened 2022-08-03
Update k8s CustomResourceDefinition

Twsouza:
Update the CRDs to match k8s 1.22.

### DIFF
--- a/provider/k8s/k8s.go
+++ b/provider/k8s/k8s.go
@@ -28,6 +28,10 @@ import (
 	metricsclientset "k8s.io/metrics/pkg/client/clientset/versioned"
 )
 
+const (
+	CURRENT_CM_VERSION = "v1.9.0"
+)
+
 type Provider struct {
 	Atom          atom.Interface
 	CertManager   bool
@@ -279,7 +283,7 @@ func (p *Provider) initializeTemplates() error {
 			return p.applySystemTemplate("cert-manager", nil)
 		}
 
-		if d.Spec.Template.Labels["app.kubernetes.io/version"] != "v1.9.0" {
+		if d.Spec.Template.Labels["app.kubernetes.io/version"] != CURRENT_CM_VERSION {
 			fmt.Println("Updating cert-manager")
 			p.deleteSystemTemplate("cert-manager", nil)
 
--- a/provider/k8s/k8s.go
+++ b/provider/k8s/k8s.go
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	CURRENT_CM_VERSION = "v1.9.0"
+	CURRENT_CM_VERSION    = "v1.9.0"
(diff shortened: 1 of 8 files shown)


provider/k8s/k8s.go
```diff
@@ -263,8 +274,28 @@ func (p *Provider) initializeTemplates() error {
...
 
 	if p.CertManager {
-		if err := p.applySystemTemplate("cert-manager", nil); err != nil {
-			return errors.WithStack(err)
+		d, _ := p.Cluster.AppsV1().Deployments("cert-manager").Get("cert-manager", am.GetOptions{})
+		if d == nil {
+			return p.applySystemTemplate("cert-manager", nil)
+		}
+
+		if d.Spec.Template.Labels["app.kubernetes.io/version"] != "v1.9.0" {
```
heronrs, line 282: We should probably use an env variable for the cert-manager version so we don't need to change it every time we update its version. For now I think it's ok though, we can do it on the next iteration.

provider/k8s/k8s.go
```diff
@@ -263,8 +274,28 @@ func (p *Provider) initializeTemplates() error {
...
+			fmt.Println("Updating cert-manager")
+			p.deleteSystemTemplate("cert-manager", nil)
+
+			for {
+				n, _ := p.Cluster.CoreV1().Namespaces().Get("cert-manager", am.GetOptions{})
+				if n.Name == "" {
```
heronrs, line 288: should we also check `if n != nil` since it's a pointer?
If `Namespaces().Get()` always return a non nil `n` it's fine though.
Also found this [check](https://github.com/convox/convox/blob/master/provider/k8s/app.go#L97) that it's a bit more idiomatic I think.

heronrs: We should also add a max retry and a backoff time.
something like
```go
maxRetries := 10
retries := 0 // increment by one on each loop iteration
for {
  ...
 time.Sleep(time.Seconds * 10)
}
```

If it doesn't delete the ns within the threshold just print the error and return

heronrs (commented):
pretty good, just some minor comments

provider/k8s/k8s.go
Twsouza: > should we also check `if n != nil` since it's a pointer? If `Namespaces().Get()` always return a non nil `n` it's fine though. Also found this [check](https://github.com/convox/convox/blob/master/provider/k8s/app.go#L97) that it's a bit more idiomatic I think.

the `Namespaces().Get()` returns an namespace object with an empty metadata, using the `n != nil` doesn't work...

heronrs: Also, for all errors inside `func (p *Provider) initializeTemplates() error ` let's just `panic` them so we still keep the same behavior and fail the API initialization.

heronrs (approved):

PR #470, opened 2022-08-05
Creating a new app use an empty Atom CRD

Twsouza:
When upgrading from 3.5.7 to 3.6.0 and trying to create a new app, the CLI tried to create an empty resource using Atom CRD.

### DIFF
--- a/pkg/atom/templates/crd.yml.tmpl
+++ b/pkg/atom/templates/crd.yml.tmpl
@@ -18,10 +18,13 @@ spec:
           properties:
             started:
               type: string
+              nullable: true
             status:
               type: string
+              nullable: true
             spec:
               type: object
+              nullable: true
               properties:
                 conditions:
                   type: array


pkg/atom/templates/crd.yml.tmpl
```diff
@@ -18,10 +18,13 @@ spec:
           properties:
             started:
               type: string
+              nullable: true
             status:
               type: string
+              nullable: true
             spec:
               type: object
+              nullable: true
```
Twsouza, line 27: In the [Apply](https://github.com/convox/convox/blob/b4650f7b0cc941fd44ebc0ca6baed956db0cd851/pkg/atom/atom.go#L138) function, on the `atom` package, was trying to create an empty resource and it was failing because in the schema the tags weren't defined as nullable.

```
                 a, err = c.Atom.AtomV1().Atoms(ns).Create(&aa.Atom{
			ObjectMeta: am.ObjectMeta{
				Name: name,
			},
		})
		if err != nil {
			return errors.WithStack(err)
		}
```

heronrs (approved):

nightfury1204 (approved):

PR #474, opened 2022-08-11
Scaling memory and cpu via convox cli not working

yuriadams:


provider/k8s/service.go
```diff
@@ -177,6 +178,23 @@ func (p *Provider) ServiceUpdate(app, name string, opts structs.ServiceUpdateOpt
 		d.Spec.Replicas = &c
 	}
 
+	if opts.Cpu != nil {
+		cpuSize := resource.MustParse(fmt.Sprintf("%dm", *opts.Cpu))
+
+		d.Spec.Template.Spec.Containers[0].Resources.
+			Requests[v1.ResourceCPU] = cpuSize
```
Twsouza, line 185: Can we add `Limits` to the CPU resource? To ensure you don't use more than you should

Twsouza (changes requested):
very good, just one note!

Twsouza (approved):
👍

PR #480, opened 2022-08-16
Change EBS CSI Driver image for arm64

Twsouza:
The arm64 image was deleted, and the image on `v1.9.0` is multi-arch.

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1315

Also fixes:

- Release notes;
- remove deprecation warning;
- `proxy_protocol` and ebs csi driver installation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - name: golang-1.16
         uses: actions/setup-go@v3
         with:
--- a/terraform/cluster/aws/main.tf
+++ b/terraform/cluster/aws/main.tf
@@ -152,7 +152,7 @@ module "ebs_csi_driver_controller" {
   version = "3.3.1"
 
   ebs_csi_controller_image                   = "public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver"
-  ebs_csi_driver_version                     = var.arm_type ? "v1.9.0-linux-arm64-amazon" : "v1.9.0"
+  ebs_csi_driver_version                     = "v1.9.0"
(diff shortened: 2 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
Twsouza, line 14: Fix the release notes, it was failing because `checkout@v3` only gets the last commit.

terraform/cluster/aws/main.tf
```diff
@@ -144,18 +144,30 @@ resource "aws_launch_template" "cluster" {
...
+  depends_on = [
+    null_resource.wait_k8s_api
+  ]
+
   source  = "DrFaust92/ebs-csi-driver/kubernetes"
   version = "3.3.1"
 
   ebs_csi_controller_image                   = "public.ecr.aws/ebs-csi-driver/aws-ebs-csi-driver"
-  ebs_csi_driver_version                     = var.arm_type ? "v1.9.0-linux-arm64-amazon" : "v1.9.0"
+  ebs_csi_driver_version                     = "v1.9.0"
```
Twsouza, line 155: `v1.9.0-linux-arm64-amazon` tag doesn't exist anymore, changing to `v1.9.0` (which is multi-arch, it will work on amd64 and arm64)

terraform/router/aws/outputs.tf
```diff
@@ -1,3 +1,3 @@
 output "endpoint" {
-  value = data.http.alias.body
+  value = data.http.alias.response_body
```
Twsouza, line 2: `body` is deprecated, using `response_body`

nightfury1204 (approved):

PR #481, opened 2022-08-19
Split resource variables

Twsouza:
Split the `${RESOURCE_NAME}_URL` into : 

```
${RESOURCE_NAME}_USER
${RESOURCE_NAME}_PASS
${RESOURCE_NAME}_HOST
${RESOURCE_NAME}_PORT
${RESOURCE_NAME}_NAME
```

### DIFF
--- a/provider/k8s/process.go
+++ b/provider/k8s/process.go
@@ -429,7 +429,6 @@ func (p *Provider) podSpecFromService(app, service, release string) (*ac.PodSpec
 
 			for _, r := range s.ResourceMap() {
 				key := strings.Join(strings.Split(r.Env, "_")[1:], "_")
-				fmt.Printf("\n\n\n========> key for %v: %v\n\n\n", r.Env, key)
 				c.Env = append(c.Env, ac.EnvVar{
 					Name: r.Env,
 					ValueFrom: &ac.EnvVarSource{
--- a/provider/k8s/process.go
+++ b/provider/k8s/process.go
@@ -428,12 +428,14 @@ func (p *Provider) podSpecFromService(app, service, release string) (*ac.PodSpec
 			c.Image = fmt.Sprintf("%s:%s.%s", repo, service, r.Build)
 
 			for _, r := range s.ResourceMap() {
+				key := strings.Join(strings.Split(r.Env, "_")[1:], "_")
+				fmt.Printf("\n\n\n========> key for %v: %v\n\n\n", r.Env, key)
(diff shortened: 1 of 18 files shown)


provider/k8s/process.go
```diff
@@ -428,12 +428,14 @@ func (p *Provider) podSpecFromService(app, service, release string) (*ac.PodSpec
 			c.Image = fmt.Sprintf("%s:%s.%s", repo, service, r.Build)
 
 			for _, r := range s.ResourceMap() {
+				key := strings.Join(strings.Split(r.Env, "_")[1:], "_")
+				fmt.Printf("\n\n\n========> key for %v: %v\n\n\n", r.Env, key)
```
nightfury1204, line 432: is this print line necessary?

yuriadams (approved):

nightfury1204 (approved):
:fire:

provider/k8s/process.go
```diff
@@ -428,12 +428,13 @@ func (p *Provider) podSpecFromService(app, service, release string) (*ac.PodSpec
 			c.Image = fmt.Sprintf("%s:%s.%s", repo, service, r.Build)
 
 			for _, r := range s.ResourceMap() {
+				key := strings.Join(strings.Split(r.Env, "_")[1:], "_")
```
beastawakens, line 431: @Twsouza This breaks things for any resource with a 'multi-part' name.
i.e. I have a resource called `primary-database` in the convox.yml.  This means I end up with a env var called `PRIMARY_DATABASE_URL`.  
But splitting on the underscore and pulling out everything after the first part, means that the `key` is now `DATABASE_URL` which doesn't exist in the ConfigMap so deployments fail as that can't be found.

provider/k8s/template.go
```diff
@@ -82,6 +82,9 @@ func (p *Provider) templateHelpers() template.FuncMap {
 		"k8sname": func(s string) string {
 			return nameFilter(s)
 		},
+		"resourceEnv": func(s string) string {
+			return strings.Join(strings.Split(s, "_")[1:], "_")
```
beastawakens, line 86: I think this will need to be changed as well.

provider/k8s/process.go
Twsouza: hey @beastawakens thanks for pointing this out. I've opened a PR to fix it, see https://github.com/convox/convox/pull/517

PR #484, opened 2022-08-23
Configurable concurrency for timers

yuriadams:
[Feature 406](https://github.com/convox/issues-private/issues/406)


Twsouza:
can you add the documentation about this parameter on `docs/reference/primitives/app/timer.md`?

yuriadams:
Sure, I've just asked in the slack channel for help. How can I add this in the docs.

docs/reference/primitives/app/timer.md
```diff
@@ -28,6 +28,7 @@ A Timer is defined in [`convox.yml`](/configuration/convox-yml).
 | **command**  | **yes**  | The command to execute once the [Process](/reference/primitives/app/process) starts                               |
 | **schedule** | **yes**  | A cron formatted schedule for spawning the [Process](/reference/primitives/app/process). All times are UTC        |
 | **service**  | **yes**  | The name of the [Service](/reference/primitives/app/service) that will be used to spawn the [Process](/reference/primitives/app/process) |
+| **concurrencyPolicy**  | **no**  | It specifies how to treat concurrent executions of a job that is created by this cron job. Check this [doc](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy) for more info. |
```
heronrs, line 31: this is a bit of a nit but could we change this param to `concurrency` only? just to keep it short just like the other parameters? Also, I think the default should be `forbid` instead of `no` (IIRC `forbid` it what we use by default in the Backend). We should probably also allow lowercase options (e.g `forbid`) as I don't think we enforce Capitalize parameters anywhere in `convox.yml`

#nit2 the MD table formatting looks a bit off

yuriadams: hey, @heronrs, in this case, `no` is not the default value but if this field is required.

heronrs: oh lol ok thanks, my bad

docs/reference/primitives/app/timer.md
```diff
@@ -25,9 +25,10 @@ A Timer is defined in [`convox.yml`](/configuration/convox-yml).
...
-| **command**  | **yes**  | The command to execute once the [Process](/reference/primitives/app/process) starts                               |
-| **schedule** | **yes**  | A cron formatted schedule for spawning the [Process](/reference/primitives/app/process). All times are UTC        |
-| **service**  | **yes**  | The name of the [Service](/reference/primitives/app/service) that will be used to spawn the [Process](/reference/primitives/app/process) |
+| **command**      | **yes**  | The command to execute once the [Process](/reference/primitives/app/process) starts                               |
+| **schedule**     | **yes**  | A cron formatted schedule for spawning the [Process](/reference/primitives/app/process). All times are UTC        |
+| **service**      | **yes**  | The name of the [Service](/reference/primitives/app/service) that will be used to spawn the [Process](/reference/primitives/app/process) |
+| **concurrency**  | **no**   | It specifies how to treat concurrent executions of a job that is created by this cron job. Check this [doc](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy) for more info. |
```
heronrs, line 31: Can you also add that when not set the default is `forbid`?

heronrs (approved):
LGTM, just a minor detail on the field description

docs/reference/primitives/app/timer.md
yuriadams: okay, I will explicit the default value. It was described in the docs but I agree with you.

heronrs (approved):

yuriadams:
@heronrs, the deepsource is complaining about the strings.Title function is deprecated. But for the version we're using, 1.17, is not. So, what do you suggest to me?

provider/k8s/release.go
```diff
@@ -470,6 +470,10 @@ func (p *Provider) releaseTemplateServices(a *structs.App, e structs.Environment
 }
 
 func (p *Provider) releaseTemplateTimer(a *structs.App, e structs.Environment, r *structs.Release, s *manifest.Service, t manifest.Timer) ([]byte, error) {
+	if t.Concurrency != "" {
```
yuriadams, line 473: @heronrs, the deepsource is complaining about this `strings.Title`. It's saying that is deprecated in go 1.18 but we're not in this version yet. What should I do here? Btw, I didn't like my solution.

heronrs (approved):

PR #485, opened 2022-08-25
E2e tests for timers v3

yuriadams:
The test will:

- Test if a default timer is running, check the pod output
- Check if the concurrency allowed job has more than 1 pod running in parallel
- Check if the concurrency forbid job has just 1 pod running


yuriadams:
This PR has https://github.com/convox/convox/pull/484 as a dependency.

nightfury1204:
Hey @yuriadams did you run the e2e tests on the github actions for your changes?
If not can you run it please?
I am facing error on `convox logs -a httpd --no-follow --since 10m | grep service/web`, because i added another service in `convox.yaml` kind of similar to yours.

examples/httpd/convox.yml
```diff
@@ -2,3 +2,21 @@ services:
...
     image: httpd
     port: 80
+  worker:
+    build: ./worker
+
+timers:
+  cleanup:
+    command: ./cmd.sh
+    schedule: "* * * * *"
+    service: worker
```
Twsouza, line 12: Rename the timer to a more descriptive (e.g. `example`) and link it to the `web` service.

Twsouza (changes requested):

examples/httpd/convox.yml
yuriadams: hey, @Twsouza. What do you mean about "link it to the web service"?

Twsouza: instead of using the `worker` on `services` for the timer, use the `web` service:

```
example:
    command: ./cmd.sh
    schedule: "* * * * *"
    service: worker
```

You may need to create a Dockerfile to the `web` service to copy the shell script.

yuriadams: hey @Twsouza, thanks for your feedback, man. I've just updated and remove the dependency from the concurrency task. Before merge it, I'll do the respective tests.

ci/test.sh
```diff
@@ -102,5 +102,13 @@ convox ps stop $ps -a httpd
 convox ps -a httpd | grep -v $ps
 convox deploy -a httpd
 
+# timers
+sleep 30
+
+timerLog=$(convox logs -a httpd --no-follow --since 1m | grep service/web/timer-example)
```
nightfury1204, line 108: Is 30 second sleeping enough?
from schedule `schedule: "* * * * *"` i can see cron should run in at every minutes
also `--since 1m` seems very strict in that sense

ci/test.sh
```diff
@@ -102,5 +102,13 @@ convox ps stop $ps -a httpd
...
+if ! [[ $timerLog == *"Hello Timer"* ]]; then
+  echo "failed"; exit 1;
+fi
+
```
Twsouza, line 112: Use the same pattern as in the other tests.

E.g.:
```suggestion
convox logs -a httpd --no-follow --since 1m | grep service/web/timer-example | grep ""Hello Timer"
```

ci/test.sh
```diff
@@ -102,5 +102,13 @@ convox ps stop $ps -a httpd
 convox ps -a httpd | grep -v $ps
 convox deploy -a httpd
 
+# timers
+sleep 30
```
Twsouza, line 106: It seems that 30 secs is not enough https://github.com/convox/convox/runs/8207383225?check_suite_focus=true
maybe using 2m?

Twsouza (changes requested):

ci/test.sh
yuriadams: hey, @nightfury1204. Yes, it has been enough to pass the check.

ci/test.sh
yuriadams: Actually @Twsouza, this is another issue that I'll fix in the next commit. 30 seconds is enough. Well, at least, so far.

Twsouza (approved):
lgtm

PR #486, opened 2022-09-01
Feat-570: Add e2e downgrade test

nightfury1204:
https://github.com/convox/issues-private/issues/570

e2e test run: https://github.com/convox/convox/runs/8122037532?check_suite_focus=true

Added postgres and redis resource testing.

For downgrate test: it first runs all the usual tests and setup the resource, services, then downgrades it to the latest release and then run all the usual tests and the existing resource and service test.

Using image and build in `convox.yml` caused weird behaviour(https://github.com/convox/issues-private/issues/594). Discussed with @heronrs to use image for both now to unblock the progress.


examples/httpd/convox.yml
```diff
@@ -1,4 +1,18 @@
...
+    type: redis
 services:
   web:
     image: httpd
     port: 80
+  resource-tester:
+    image: nightfury1204/resource-tester:0.0.1
+    resources:
+      - postgresdb:DB_URL
+      - rediscache:REDIS_URL
```
Twsouza, line 14: Attach the resources to the `web` service, don't need to create an extra service for this.

It's better to use the build option instead of a pre-built image for tests.

```
image: nightfury1204/resource-tester:0.0.1
```

Twsouza (changes requested):

examples/httpd/convox.yml
nightfury1204: I did went for  for the build option, but there is a issue: https://github.com/convox/issues-private/issues/594

I went for multi service so that it will be tested for multi service setup also

.github/workflows/release.yml
```diff
@@ -147,8 +152,18 @@ jobs:
...
         run: sleep 600
+      - name: before downgrade test
+        run: ci/test.sh downgrade-setup
+        if: ${{ matrix.downgrade == true }}
+      - name: downgrade
+        run: ci/downgrade.sh
+        if: ${{ matrix.downgrade == true }}
+      - name: after downgrade test
+        run: ci/test.sh downgrade-check
+        if: ${{ matrix.downgrade == true }}
```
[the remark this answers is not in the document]
nightfury1204: Since in downgrade you go from one version to another, that's why i think it's necessary to run test after the downgrade.
what do you think?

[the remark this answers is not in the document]
Twsouza: Yeah, I agree (that is what I was trying to say 😅, I think I wasn't clear) only test after the downgrade and not before, since before it will be tested in the `ci (aws)` job.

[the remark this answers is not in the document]
nightfury1204: sorry got it now, yeah i agree before test actually not necessary

examples/httpd/convox.yml
Twsouza: Okay, great job catching this up 👍 

About the image, I don't think using your docker hub repository is a good option, maybe @heronrs has access to the convox repo on docker hub?

.github/workflows/release.yml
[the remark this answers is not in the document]
Twsouza: I'm changing v2 to have the same behavior 👷‍♂️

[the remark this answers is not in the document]
nightfury1204: addressed it

ci/test.sh
```diff
@@ -33,6 +33,49 @@ convox rack logs --no-follow --since 30m
...
+    convox exec $ps "/usr/scripts/db_insert.sh" -a httpd2
+    convox exec $ps "/usr/scripts/db_check.sh" -a httpd2
+    # redis resource test
+    convox resources -a httpd2 | grep rediscache
+    ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "resource-tester")|.id' | head -n 1)
+    convox exec $ps "/usr/scripts/redis_check.sh" -a httpd2
+
+    # downgrate-setup runs before the downgrading, so no need to test others functionality
+    # they are covered by other ci test
+    exit 0
```
Twsouza, line 60: Creating another shell script will make this clear.

.github/workflows/release.yml
Twsouza, line 163: Create a different script for each step, it's more readable.

ci/downgrade.sh
```diff
@@ -0,0 +1,23 @@
...
+while [ "${status}" != "running" ]
+do
+    echo "waiting for rack updates to complete..."
+    sleep 30
+    status=$(convox rack | grep 'Status' | xargs | cut -d ' ' -f2)
+    echo "rack status: ${status}"
+    if [ "${status}" = "failed" ]; then
+        exit 1
+    fi
+done
```
Twsouza, line 23: Do we need to wait until the rack returns "running" status? The V3 CLI waits until the update it's completed by default.

ci/test.sh
```diff
@@ -33,6 +33,49 @@ convox rack logs --no-follow --since 30m
...
+    # postgres resource test
+    ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "resource-tester")|.id' | head -n 1)
+    convox exec $ps "/usr/scripts/db_check.sh" -a httpd2
+    # redis resource test
+    convox resources -a httpd2 | grep rediscache
+    ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "resource-tester")|.id' | head -n 1)
+    convox exec $ps "/usr/scripts/redis_check.sh" -a httpd2
+    #cleanup
+    convox apps delete httpd2
+fi
```
Twsouza, line 77: Move it to another script and reference it in the YAML config.

ci/test.sh
```diff
@@ -72,17 +115,28 @@ endpoint=$(convox api get /apps/httpd/services | jq -r '.[] | select(.name == "w
...
+i=0
+while [ "$(convox ps -a httpd | grep web | wc -l)" != "2" ]
+do
+    if [ $((i++)) -gt 50 ]; then
+        exit 1
+    fi
+    echo "waiting for web to scale up..."
+    convox ps -a httpd
+    sleep 15
+done
```
Twsouza, line 138: The V3 CLI wait until the scale is complete before returning, doesn't it?
If necessary you can decrease the counter number, the current `50` It can take up to 13 minutes to return a fail.

Twsouza (changes requested):

ci/test.sh
heronrs: Using a different shell script might be a good option since we'll also be adding upgrade-setup and upgrade-check

.github/workflows/release.yml
```diff
@@ -114,6 +115,10 @@ jobs:
             provider: aws
             rack_params: "node_type=t4g.small proxy_protocol=true"
             suffix: arm64
+          - job: aws-downgrade
```
heronrs, line 118: Did you check about the manual approval for this job?
You can implement it in a different ticket if you want, let's just track it in the milestone.

heronrs (commented):
Looks pretty good to me other than Taynan's comments.
Also, check my comment about the manual approval setup.

ci/test.sh
nightfury1204: This one is necessary probably in gcp provider sometimes it takes a little more time.
So i am going to reduce it to 10.

.github/workflows/release.yml
nightfury1204: Created another ticket for that: https://github.com/convox/issues-private/issues/606

Twsouza (changes requested):
@nightfury1204 can you look at the `ci (aws-existing-vpc)` test? It's failing with `Process completed with exit code 2.`

nightfury1204:
> @nightfury1204 can you look at the `ci (aws-existing-vpc)` test? It's failing with `Process completed with exit code 2.`

@Twsouza it was network issue with ecr registry, retrying it worked: https://github.com/convox/convox/runs/8227274206?check_suite_focus=true

heronrs (dismissed):
:100:

heronrs (approved):

heronrs:
@Twsouza I'm merging this one. Let us know if you think something else should be changed.

PR #491, opened 2022-09-13
feat-606: Add manual approval in e2e test ci

nightfury1204:
https://github.com/convox/issues-private/issues/606

ci : https://github.com/convox/convox/actions/runs/3045546296

https://user-images.githubusercontent.com/12232198/189977609-ce6d6bd6-3d1c-49cb-aa5d-4a519fdbf2cc.mp4

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -102,9 +102,7 @@ jobs:
         job:
           - aws
           - aws-arm64
-          - aws-downgrade
           - aws-existing-vpc
-          - aws-upgrade
           - do
           - gcp
           # - azure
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -155,24 +145,94 @@ jobs:
...
-      - name: before version upgrade/downgrade setup and test
+      - name: test
+        run: ci/test.sh
+      - name: debug
+        run: ci/debug.sh
+        if: always()
+      - name: uninstall
+        run: ci/uninstall.sh
+        if: always()
+  ci-downgrage:
```
Twsouza, line 158:
```suggestion
  ci-downgrade:
```

.github/workflows/release.yml
```diff
@@ -155,24 +145,94 @@ jobs:
...
+    if: github.repository == 'convox/convox'
+    environment: 'UpgradeDowngradeTests'
```
Twsouza, line 160: Can `UpgradeDowngradeTests` be renamed? In case we want to add more additional tests...

nightfury1204: this environment is configured by @heronrs, i don't have access. He can give more context on it

.github/workflows/release.yml
nightfury1204: nice catch. fixed it

Twsouza (approved):
👍

.github/workflows/release.yml
heronrs: Yeah sorry, that was a bad name. I can't rename the environment sadly but I'll create a new one later. Maybe name it just ManuApprovalTests.

Merging for now just so we can move forward

PR #492, opened 2022-09-14
Adding in whitelist option at the service level

beastawakens:
PR incoming for `rack` as well to maintain parity to v2.

### DIFF
--- a/docs/reference/primitives/app/service.md
+++ b/docs/reference/primitives/app/service.md
@@ -67,6 +67,7 @@ services:
     timeout: 180
     tls:
       redirect: true
+    whitelist: 10.0.0.128/16,192.168.0.1/32
 ```
 
 | Attribute     | Type       | Default             | Description                                                                                                                                |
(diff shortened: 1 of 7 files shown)


heronrs:
@beastawakens this is only for apps that define their own load balancer, right?

beastawakens:
@heronrs no, the whitelist option on custom balancers already exists.  This change allows you to add a whitelist on the service level ingress object itself so it's easier!  You don't need a separate balancer if you just want to whitelist 😄

beastawakens:
Is this all good to merge @heronrs ? 😄

beastawakens:
@heronrs finally remembered to fix the tests 😆

NilsJPWerner:
@beastawakens this is incredible! Does this require proxy_protocol to be turned on?

beastawakens:
> @beastawakens this is incredible! Does this require proxy_protocol to be turned on?

It does, otherwise the IP coming through is just an internal one so makes it kinda pointless 😄 But once that's on, it works great!

PR #494, opened 2022-09-15
Add e2e tests for app intercomunication

yuriadams:
[task 613](https://github.com/convox/issues-private/issues/613)


examples/httpd/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM httpd
 
 RUN apt update > /dev/null 2>&1
-RUN apt install -y postgresql-client redis > /dev/null 2>&1
+RUN apt install -y postgresql-client redis > /dev/null 2>&1 curl > /dev/null
```
nightfury1204, line 4: put the along with postgresql and other tools like the following:
```
RUN apt install -y curl postgresql-client redis > /dev/null 2>&1
```

nightfury1204 (commented):
to me looks like we can add this test to the `test.sh` file
@yuriadams is there any reason to use separate file?

heronrs:
> to me looks like we can add this test to the `test.sh` file @yuriadams is there any reason to use separate file?

that's a good point. I think we split the setup for before/after upgrade/downgrade but maybe this type of test could be included in test.sh. Otherwise, we can split them into different files but we'll be adding a bunch of additional CI steps for every new test.
We should also run for all cases(the release job and the update/downgrade jobs) - including in test.sh avoids adding a new step to each one.

yuriadams:
> to me looks like we can add this test to the `test.sh` file @yuriadams is there any reason to use separate file?

No reason, @nightfury1204, I'm going to change it.

Twsouza (approved):
🔥

nightfury1204 (approved):

PR #495, opened 2022-09-15
e2e test for cache and db resource

nightfury1204:
https://github.com/convox/issues-private/issues/575
https://github.com/convox/issues-private/issues/576

ci run: https://github.com/convox/convox/actions/runs/3059918639

### DIFF
--- a/ci/after_additional_test.sh
+++ b/ci/after_additional_test.sh
@@ -28,12 +28,26 @@ convox apps | grep httpd2
 convox apps info httpd2 | grep running
 endpoint=$(convox env get ENDPOINT -a httpd2)
 fetch https://$endpoint | grep "It works"
+
 # postgres resource test
 ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
 convox exec $ps "/usr/scripts/db_check.sh" -a httpd2
+
+# mariadb resource test
+ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
+convox exec $ps "/usr/scripts/mariadb_check.sh" -a httpd2
+
+# mysqldb resource test
+ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
+convox exec $ps "/usr/scripts/mariadb_check.sh mysql" -a httpd2
+
 # redis resource test
-convox resources -a httpd2 | grep rediscache
 ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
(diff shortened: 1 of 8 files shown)


yuriadams (approved):

ci/after_additional_test.sh
```diff
@@ -28,12 +28,26 @@ convox apps | grep httpd2
 convox apps info httpd2 | grep running
 endpoint=$(convox env get ENDPOINT -a httpd2)
 fetch https://$endpoint | grep "It works"
+
 # postgres resource test
 ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
 convox exec $ps "/usr/scripts/db_check.sh" -a httpd2
+
+# mariadb resource test
+ps=$(convox api get /apps/httpd2/processes | jq -r '.[]|select(.status=="running" and .name == "web")|.id' | head -n 1)
```
heronrs, line 37: do we need to test these resources in the after/before shell files?
otherwise, i'd say let's just add it to `test.sh`

nightfury1204: it's for the resource persistence testing. i think other dbs also should be included for this type of test.
that's why i added it.
What do you think?

examples/httpd/convox.yml
```diff
@@ -3,14 +3,26 @@ resources:
...
+  memcache:
+    type: memcached
+    options:
+      version: 1.6.17
+  mysqldb:
+    type: mysql
+    options:
+      version: 8.0.30
+  mariadb:
+    type: mariadb
```
Twsouza, line 15: Do we need to test this specific version? Should we test using the latest?

nightfury1204: otherwise aws arm e2e fails
default version doesn't support arm architecture

ci/after_additional_test.sh
Twsouza: I think it's to check if the resource is available, right?

Twsouza (dismissed):
lgtm

ci/after_additional_test.sh
nightfury1204: yes

examples/httpd/convox.yml
Twsouza: @nightfury1204 can you open an issue to update the default version?
e.g. [mariadb](https://github.com/convox/convox/blob/775fd9b00bb7683acf184fcccca5790278d60d89/provider/k8s/template/resource/mariadb.yml.tmpl)

Twsouza: i.e. [mysql version](https://github.com/convox/convox/blob/ddbd5df1c6c22e357741c43664264346793b1795/provider/k8s/template/resource/mysql.yml.tmpl#L65)

nightfury1204: already created one for that: https://github.com/convox/issues-private/issues/621

heronrs (approved):
LGTM

PR #499, opened 2022-09-28
Feat-542: Add ssh capability for v3 rack

nightfury1204:
https://github.com/convox/issues-private/issues/542


pkg/cli/instances.go
```diff
@@ -53,16 +56,50 @@ func Instances(rack sdk.Interface, c *stdcli.Context) error {
...
+		if m.State != nil {
+			private, public, err := generateSShKeyPair()
+			if err != nil {
+				return err
+			}
+
+			c.Args = append(c.Args, fmt.Sprintf("ssh_private_key=%s", base64.StdEncoding.EncodeToString(private)))
+			c.Args = append(c.Args, fmt.Sprintf("ssh_public_key=%s", base64.StdEncoding.EncodeToString(public)))
+			c.Writef("\n")
+			return RackParamsSet(r, c)
```
heronrs, line 92: This will display the ssh* parameter when doing `convox rack params`, right?
Also, I think it will store the ssh keys unencrypted in Dynamo(the console storage).
I think instead of using a rack parameter we could directly call [` r.InstanceKeyroll()`,](https://github.com/convox/convox/blob/feat-542_ssh/provider/k8s/instance.go#L19) passing the private,public keys and set it directly in k8s, what do you think?

Then in tf we also remove the ssh public/private key and generate one in the first install/update.
In v2 racks we also don't allow users to set a custom key so I think it's ok to not support that either.

In the future, if that need arises we can come up with a way of passing custom keys via `convox instances keyroll`

nightfury1204: since the node instances are managed by terraform through launch configuration, so we have to pass it to terraform so that it can install ssh key in the node instances
that's why it is like this @heronrs

nightfury1204: But i will think about the concern you raised

heronrs: Oooh, I see what you mean.
I think a major concern should be to not display the keys unencrypted anywhere(console, rack params, Dynamo, terraform).
Instead of setting a param one idea is to instruct terraform to recreate the ssh key/pair resource

nightfury1204: recreating is the problem here.
I wanted to use https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key
but couldn't find way to recreate it when the `keyroll` cmd is run.
Do you have any idea?

heronrs: Sadly looks like https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key is not very good for production environments as it stores the key unencrypted in tf state.

Do you think running a sh script could help on creating the necessary input to aws_key_pair?
We can use [null](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) resources to run arbitrary shell scripts 
I believe in combination with [local-exec](https://www.terraform.io/language/resources/provisioners/local-exec) we can run a sh to generate the input we need for aws_key_pair and read with [file](https://www.terraform.io/language/functions/file) .

With the current console version of terraform we can [`taint` ](https://www.terraform.io/cli/v1.1.x/commands/taint) a resource and in the next apply it will recreate the resource.

And for initiating terraform I think keyroll() would need to trigger a tf run somehow(like a regular update but instead of new parameters, apply a taint to the ssk key).

I'm not sure how much work would be involved in this last part.

nightfury1204: Sounds good, let me try these

nightfury1204: Hey I tried the first approach the problem is when installing it's ok but when we are updating it showing error like the ssh key file not found. then i tried usin script running same issue either file not found or i have to create new ssh key pair in every change.

For second one using, https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key, i found the 
replace by trigger option: https://www.terraform.io/language/meta-arguments/lifecycle#replace_triggered_by
But it need updated terraform version

The keyroll() using tainting looks very troublesome, still haven't tried it.

@heronrs what should i do in these case?

nightfury1204:
https://user-images.githubusercontent.com/12232198/194627934-6419ccad-5c91-4bc5-a35d-87bfad8008be.mp4

nightfury1204:
e2e tests: https://github.com/convox/convox/actions/runs/3206321965

pkg/cli/instances.go
```diff
@@ -91,6 +111,14 @@ func InstancesSsh(rack sdk.Interface, c *stdcli.Context) error {
 		opts.Command = options.String(command)
 	}
 
+	if key := c.String("key"); key != "" {
```
heronrs, line 114: Maybe check if it's a v3 rack before looking for the key?

provider/k8s/instance_test.go
```diff
@@ -35,3 +49,61 @@ func nodeCreateResources(c kubernetes.Interface, name, cpu, mem string) error {
 		}
 	})
 }
+
+func TestInstanceShell(t *testing.T) {
```
heronrs, line 53: I'd rename this `TestInstanceShellError`  since you are asserting the `InstanceShell()` call for errors(I'm guessing bc we can't mock the `ssh.Dial` call.
For the "happy path" case I think it's worth adding an e2e test with keyroll and ssh if possible

heronrs:
@nightfury1204 forgot to put on the review, but please add into the documentation this new `key_pair_name` parameter, and document the instances keyroll and ssh commands in this section https://docs.convox.com/reference/cli/instances/

pkg/cli/instances.go
nightfury1204: added the check

provider/k8s/instance_test.go
nightfury1204: Changed the name

Twsouza (dismissed):
good job!

heronrs (dismissed):
LGTM, add the e2e tests if you can

heronrs (approved):

PR #501, opened 2022-09-30
Replace docker with buildkit

heronrs:
As we all know, k8s is removing the docker shim and since we depend on docker for our builds we also need to change our code.
This PR will replace `docker` by a daemonless `buildkit` container when running builds.
The process of import/export apps was also heavily dependent on docker. I replaced the docker part with `skopeo`, a standalone cli for inspecting/downloading/syncing docker images.


.github/workflows/release.yml
```diff
@@ -86,6 +86,10 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       - name: build-amd64
         run: "docker buildx build --platform linux/amd64 -t ${{ github.repository }}:${{ env.VERSION }} --push --cache-from type=registry,ref=${{ github.repository }} --cache-to type=registry,ref=${{ github.repository }},mode=max ."
+      - name: buildkit-amd64
```
heronrs, line 89: buildkit requires an specific image and come in two "flavors", "privileged" and "rootless". The preferred flavor should be "rootless" but there are some k8s environments that do not work with it, so we use the "privileged" version

Dockerfile
```diff
@@ -43,12 +43,12 @@ RUN make package build compress
 
 ## production ##################################################################
 
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
heronrs, line 46: [Skopeo](https://github.com/containers/skopeo/blob/main/install.md) doesn't provide a direct binary to download. Its standard distribution method is via package managers and is only available in ubuntu 20.10+. I bumped to 22 since it's the newest version, and based on our tests, no bugs were introduced. Better to keep stuff updated

Dockerfile-buildkit
```diff
@@ -0,0 +1,44 @@
+FROM golang:1.16-alpine AS package
```
heronrs, line 1: We need to embed the rack build binary into the buildkit image. The other way around doesn't work.

ci/test.sh
```diff
@@ -154,6 +159,21 @@ $root/ci/test/resources_memcache.sh &
 $root/ci/test/app_internal_communication.sh &
 wait
 
+# import/export test
+convox apps create httpd-export
+convox deploy -a httpd-export --manifest convox-export.yml
```
heronrs, line 164: I'm using a new manifest since Redis and Memcached cannot be exported.
MariaDB and Mysql are not working either(also in previous versions). I'll create a ticket to address this problem in the future.

provider/k8s/build.go
```diff
@@ -159,74 +171,62 @@ func (p *Provider) BuildExport(app, id string, w io.Writer) error {
...
-		if remote {
-			if err := exec.Command("docker", "pull", from).Run(); err != nil {
-				return errors.WithStack(err)
-			}
-		}
-
-		to := fmt.Sprintf("registry:%s.%s", service, build.Id)
-
-		if err := exec.Command("docker", "tag", from, to).Run(); err != nil {
+		if err := exec.Command("skopeo", "copy", "--src-creds", fmt.Sprintf("%s:%s", user, pass), from, to).Run(); err != nil {
```
heronrs, line 193: export the layers into a tar archive with layers on the oci format.

.github/workflows/release.yml
```diff
@@ -86,6 +86,10 @@ jobs:
...
+        run: docker buildx build -f Dockerfile-buildkit --target rootless --platform linux/amd64 -t ${{ github.repository }}:${{ env.VERSION }}-build --push --cache-from type=registry,ref=${{ github.repository }} --cache-to type=registry,ref=${{ github.repository }},mode=max .
+      - name: buildkit-amd64-privileged
+        run: docker buildx build -f Dockerfile-buildkit --target privileged --platform linux/amd64 -t ${{ github.repository }}:${{ env.VERSION }}-build-privileged --push --cache-from type=registry,ref=${{ github.repository }} --cache-to type=registry,ref=${{ github.repository }},mode=max .
```
Twsouza, line 92: What's the difference between `buildkit-amd64` and `buildkit-amd64-privileged`?

buildctl-daemonless.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
```
Twsouza, line 1: is it possible to move this script to another folder? example `/scripts`?

pkg/build/buildkit.go
```diff
@@ -0,0 +1,240 @@
...
+type BuildKit struct{}
+
+func (bk *BuildKit) Build(bb *Build, dir string) error {
+	config := filepath.Join(dir, bb.Manifest)
+
+	if _, err := os.Stat(config); os.IsNotExist(err) {
+		return fmt.Errorf("no such file: %s", bb.Manifest)
+	}
+
+	data, err := ioutil.ReadFile(config)
```
Twsouza, line 30:
```suggestion
	data, err := io.ReadFile(config)
```

pkg/build/buildkit.go
```diff
@@ -0,0 +1,240 @@
...
+	err = os.WriteFile(fmt.Sprintf("%s/.docker/config.json", os.Getenv("HOME")), f, 0600)
+	if err != nil {
+		return errors.WithStack(fmt.Errorf("failed to create registry credentials file - %s", err.Error()))
+	}
+
+	return nil
+}
+
+func (*BuildKit) cacheProvider(provider string) bool {
+	return provider != "" && strings.Contains("do az", provider)
```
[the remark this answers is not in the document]
Twsouza: Do we have documentation about the cache for builds? Should we add that?

provider/do/build.go
```diff
@@ -51,25 +41,3 @@ func (p *Provider) BuildLogs(app, id string, opts structs.LogsOptions) (io.ReadC
...
-	cmd := exec.Command("docker", "login", "-u", user, "--password-stdin", repo)
-
-	cmd.Stdin = strings.NewReader(pass)
-
-	if err := cmd.Run(); err != nil {
-		return err
-	}
-
-	return nil
-}
```
Twsouza, line 75: Is it used to log into 3rd party repositories? Is it going to be needed with BuildKit?

Twsouza (commented):
🚀

provider/do/build.go
heronrs: :thinking: that's a good point. I'll do some tests

.github/workflows/release.yml
heronrs: the `privileged` variation uses a different target `--target privileged`. The privileged image is a bit less secure than the ` buildkit-amd64` because it runs as root and need some additional escalation privileges.
I built both because the preferred one should be `buildkit-amd64` but not all cloud providers support that version, probably because of differences in the underlying work nodes configuration.
`privileged` is just the standard nomenclature in Buildkit's repostiory

yuriadams (approved):

Dockerfile-buildkit
Twsouza, line 1: nit pick: @heronrs could you rename it to `Dockerfile.buildkit` so IntelliSense and code formatted understand that this is a Dockerfile?

docs/reference/primitives/app/build.md
```diff
@@ -51,20 +58,31 @@ You can define the location to build for each [Service](/reference/primitives/ap
...
+    convox builds import -a myapp2 -f /tmp/build.tgz
 ```
-> Importing a Build creates a new [Release](/reference/primitives/app/release) that references the Build.
\ No newline at end of file
+
+> Importing a Build creates a new [Release](/reference/primitives/app/release) that references the Build.
+
+### Build Layers Caching
+
+From version 3.11.0 onward, the rack uses buildkit to build and push images. Buildkit allows us to specify a caching path in remote repositories to store/fetch layers that have already been created. Unfortunately, the only rack registries that support such feature so far are DigitalOcean(DO racks have a build-in registry) and Azure.
```
Twsouza, line 88: nitpicks: 

```
From version 3.11.0 onward, Convox uses buildkit to build and push images.
```

```
(DO racks have a built-in registry)
```

provider/do/build.go
Twsouza: it's already being handled here https://github.com/convox/convox/blob/419d89bc28a8375bdab6eb725cb49d063376234b/provider/k8s/build.go#L193

provider/k8s/process.go
```diff
@@ -314,6 +323,7 @@ func (p *Provider) ProcessRun(app, service string, opts structs.ProcessRunOption
 
 	ps, err := p.ProcessGet(app, pd.ObjectMeta.Name)
 	if err != nil {
+		println("error creating build pod", err.Error())
```
Twsouza, line 326: nitpick: you may have forgotten to delete it 🤔

vendor/github.com/convox/stdsdk/client.go
```diff
@@ -407,6 +406,7 @@ func unmarshalReader(r io.ReadCloser, out interface{}) error {
 	}
 
 	data, err := ioutil.ReadAll(r)
+
```
Twsouza, line 409: can you remove this file change? so it will not show if someone runs `go vendor mod`.

Dockerfile.buildkit
```diff
@@ -0,0 +1,44 @@
+FROM golang:1.16-alpine AS package
```
Twsouza, line 1: is it possible to rename the file to `buildkit.Dockerfile`?
so any IDE/code editor would recognize this as a Dockerfile and we would have the intellisense, format and everything?

vendor/github.com/convox/stdsdk/client.go
```diff
@@ -305,7 +305,6 @@ func (c *Client) Request(method, path string, opts RequestOptions) (*http.Reques
 	}
 
 	endpoint := fmt.Sprintf("%s://%s%s%s?%s", c.Endpoint.Scheme, c.Endpoint.Host, c.Endpoint.Path, path, qs)
-
```
Twsouza, line 308: can you discard this change?

Twsouza (changes requested):
awesome!
just two small comments

Dockerfile.buildkit
heronrs: Could you let me know if you are sure about this?
You asked me to rename it to `Dockerfile.buildkit` in a previous comment :smile: and after that VS Code is picking up the file format correctly.

Twsouza: oh sorry, I got confused, you're right. the right pattern is `Dockerfile.somename`
thanks

Twsouza (dismissed):

terraform/api/azure/variables.tf
```diff
@@ -1,3 +1,7 @@
+variable "buildkit_enabled" {
+  default = false
```
Twsouza, line 2: shouldn't this be true for Azure? See https://github.com/convox/convox/pull/564

Twsouza: @nightfury1204

Twsouza (changes requested):
@nightfury1204 see my last review

Twsouza (approved):

yuriadams (approved):

PR #508, opened 2022-10-10
Node group should be able to scale in non HA mode

heronrs:
Just like v2, even when in non-HA mode, the node group should still be able to scale to accommodate whatever workloads are being scheduled.


Twsouza (approved):
:shipit:

nightfury1204 (approved):

PR #472, opened 2022-08-09
Upgrade Nginx ingress controller from v0.48 to the latest version. 1.3.0

yuriadams:

PR #510, opened 2022-10-11
Update Nginx controller to 1.3.0

Twsouza:
### What is the feature/fix?

Update the nginx ingress controller to 1.3.0, see https://github.com/convox/convox/pull/472. 
Include a tag to watch ingresses without class for old apps. When the app is deployed, it will include it.

No, old apps are still compatible and new apps will be compatible with old racks (if a downgrade is necessary).

### How to use/test it?

1. Install a rack setting release as [nginx-update-130-rc](https://github.com/convox/convox/releases/tag/nginx-update-130-rc). 
2. Create and deploy a new app. 
3. Get access to Kubernetes 
4. Search for the apps ingress `kubectl get ingress web -n dev-v3-nodejs -o yaml`
5. Under spec, it must have the `ingressClassName` as `nginx`:

```
[...]
spec:
  ingressClassName: nginx
  rules:
  - host: xxxx
[...]
```

### Checklist
- [x] New coverage tests 
**don't need, it's just a lib update**
- [x] Unit tests passing
- [x] E2E tests passing
- [x] Documentation updated

### DIFF
--- a/terraform/router/nginx/main.tf
+++ b/terraform/router/nginx/main.tf
@@ -34,6 +34,16 @@ resource "kubernetes_service_account" "ingress-nginx" {
   }
 }
 
+resource "kubernetes_ingress_class" "nginx" {
+  metadata {
+    name = "nginx"
+  }
+
+  spec {
+    controller = "k8s.io/ingress-nginx"
+  }
+}
+
 resource "kubernetes_cluster_role" "ingress-nginx" {
   metadata {
     name = "ingress-nginx"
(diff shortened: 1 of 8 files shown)


terraform/router/nginx/main.tf
```diff
@@ -189,11 +231,14 @@ resource "kubernetes_deployment" "ingress-nginx" {
           image = var.nginx_image
           args = [
             "/nginx-ingress-controller",
+            "--watch-ingress-without-class=true",
```
heronrs, line 234: :fire:

heronrs (approved):

PR #511, opened 2022-10-18
Feat-651: Add wildcard domain flag for v2 racks

nightfury1204:
### What is the feature/fix?
https://github.com/convox/issues-private/issues/651

Add a wildcard flag for wildcard domain for review app

Check details in below PR:
https://github.com/convox/rack/pull/3598

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/structs/build.go
+++ b/pkg/structs/build.go
@@ -26,11 +26,12 @@ type Build struct {
 type Builds []Build
 
 type BuildCreateOptions struct {
-	Description *string `flag:"description,d" param:"description"`
-	Development *bool   `flag:"development" param:"development"`
-	External    *bool   `flag:"external" param:"external"`
-	Manifest    *string `flag:"manifest,m" param:"manifest"`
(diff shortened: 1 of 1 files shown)


heronrs (commented):
Do we need this param even though it's only available for v2 racks?
I guess if we don't add it, it will break marshaling of the  buildCreate API call?

nightfury1204:
> Do we need this param even though it's only available for v2 racks? I guess if we don't add it, it will break marshaling of the buildCreate API call?

It's necessary as you mentioned for the buildCreate API call

nightfury1204:
> @nightfury1204 you will need to add the param to the [action-build](https://github.com/convox/action-build) action.

@Twsouza sorry why do we need it there?

Twsouza:
> @Twsouza sorry why do we need it there?

oh sorry, this param will only be used by the console to create the review app?
Some users use GitHub actions for their CI process, however, they can't create review apps in this case.

PR #512, opened 2022-10-21
Preparation for k8s 1.22 update

heronrs:
### What is the feature/fix?

This will update some k8s deprecated ingress resources we use in terraform and also update the metrics server deployment.
I also made some tweaks on the nginx deployment as the current config has some [known](https://github.com/kubernetes/ingress-nginx/issues/4061#issuecomment-1118439600) bugs.

Nope

### How to use/test it?

The current e2e tests should cover it, as it communicated with the API server.

### Checklist
- [X] New coverage tests - not needed, it's a simple tf change
- [X] Unit tests passing
- [X] E2E tests passing
- [X] E2E downgrade/update test passing
- [ ] Documentation updated
- [X] No warnings or errors on Deepsource/Codecov

### DIFF
--- a/terraform/api/k8s/main.tf
+++ b/terraform/api/k8s/main.tf
@@ -289,6 +289,7 @@ resource "kubernetes_ingress_v1" "api" {
   }
 
   spec {
+    ingress_class_name = "nginx"
     tls {
       hosts       = ["api.${var.domain}"]
       secret_name = "api-certificate"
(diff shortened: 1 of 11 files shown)


terraform/api/k8s/main.tf
```diff
@@ -268,7 +268,7 @@ resource "kubernetes_service" "api" {
   }
 }
 
-resource "kubernetes_ingress" "api" {
+resource "kubernetes_ingress_v1" "api" {
```
heronrs, line 271: If we do this type of change while also updating to k8s 1.22, terraform will fail to deleted the previous ingress resource bc the resource spec was removed in 1.22

heronrs (commented):
Clarification

yuriadams (approved):

Twsouza (approved):
🔥

PR #517, opened 2022-10-31
Return only the suffix for additional credential

Twsouza:
### What is the feature/fix?

Fix multi-name resources, it will return only the additional env var suffix. As described [here](https://github.com/convox/convox/pull/481#discussion_r1008209750), apps linking to resources with multi-part names are failing to boot because it's referencing the wrong data in the ConfigMap. It will get only the `URL`, `USER`, `PASS`, ...

No, it's a fix.

### How to use/test it?

1. On a V3 rack, deploy an app with a service linking to a multi-part name resource:

```
resources:
  primary-database:
    type: postgres
    options:
      storage: 10
services:
  web:
    build: .
    port: 3000
    resources:
      - primary-database
```

### Checklist
- [x] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [x] Documentation updated

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -96,6 +96,12 @@ convox exec -a httpd $ps -- env | grep "ONE_PASS="
 convox exec -a httpd $ps -- env | grep "ONE_HOST="
 convox exec -a httpd $ps -- env | grep "ONE_PORT="
 convox exec -a httpd $ps -- env | grep "ONE_NAME="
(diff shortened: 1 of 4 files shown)


beastawakens (approved):
🚢

yuriadams (approved):

nightfury1204 (approved):
:fire:

PR #525, opened 2022-11-02
Proxy protocol default is false

Twsouza:
### What is the feature/fix?

Update documentation about proxy protocol default value.


yuriadams (approved):

heronrs (approved):
thanks

PR #527, opened 2022-11-03
Use indirect depends_on EKS Cluster to avoid error cycle

Twsouza:
### What is the feature/fix?

The `depends_on` flag in the outputs from the cluster module is causing the dependency cycle. Although `depends_on` on outputs should be avoided it's needed to avoid timeout errors when creating the kubernetes_services/kubernetes_ingress.

> The depends_on argument should be used only as a last resort. When using it, always include a comment explaining why it is being used, to help future maintainers understand the purpose of the additional dependency.

See [documentation](https://developer.hashicorp.com/terraform/language/values/outputs#depends_on-explicit-output-dependencies).

No.

### How to use/test it?

You'd need changes on any Kubernetes resources to test it.

### Checklist
- [x] ~~New coverage tests~~
- [x] Unit tests passing
- [x] E2E tests passing
- [x] ~~Documentation updated~~

### DIFF
--- a/terraform/cluster/aws/main.tf
+++ b/terraform/cluster/aws/main.tf
@@ -46,7 +46,6 @@ resource "null_resource" "wait_k8s_api" {
   depends_on = [
     aws_eks_cluster.cluster
   ]
-
 }
 
 resource "aws_eks_cluster" "cluster" {
@@ -118,6 +117,16 @@ resource "aws_eks_node_group" "cluster" {
     create_before_destroy = true
     ignore_changes        = [scaling_config[0].desired_size]
   }
+}
+
+resource "null_resource" "wait_k8s_cluster" {
+  provisioner "local-exec" {
+    command = "sleep 10"
+  }
+
+  depends_on = [
+    aws_eks_node_group.cluster
+  ]
 }
 
 resource "local_file" "kubeconfig" {
--- a/terraform/cluster/do/outputs.tf
+++ b/terraform/cluster/do/outputs.tf
@@ -1,24 +1,19 @@
 output "ca" {
-  depends_on = [digitalocean_kubernetes_cluster.rack]
-  value      = base64decode(digitalocean_kubernetes_cluster.rack.kube_config[0].cluster_ca_certificate)
+  value = base64decode(digitalocean_kubernetes_cluster.rack.kube_config[0].cluster_ca_certificate)
 }
 
 output "endpoint" {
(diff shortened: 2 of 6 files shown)


yuriadams (dismissed):

terraform/cluster/aws/main.tf
```diff
@@ -46,7 +46,16 @@ resource "null_resource" "wait_k8s_api" {
...
     aws_eks_cluster.cluster
   ]
+}
+
+resource "null_resource" "wait_k8s_cluster" {
+  provisioner "local-exec" {
+    command = "sleep 10"
+  }
 
+  depends_on = [
```
heronrs, line 56: should this be `aws_eks_node_group.cluster` ?

Twsouza: thanks for pointing it out, I've fixed the other providers

terraform/cluster/do/outputs.tf
```diff
@@ -1,24 +1,24 @@
 output "ca" {
-  depends_on = [digitalocean_kubernetes_cluster.rack]
+  depends_on = [ null_resource.wait_k8s_cluster ]
```
heronrs, line 2: Might be wrong but I think the only one that will need a null resource is the AWS install because the output value is from a different resource than it had on the "depends_on" clause.
All others seem to use the same resource as the depends on clause - just removing depends_on will create an implicit dependency.

Twsouza: I haven't tested it on DO/Azure/GCP, it's working on AWS.

terraform/cluster/do/outputs.tf
```diff
@@ -1,24 +1,19 @@
...
 }
 
 output "name" {
-  depends_on = [digitalocean_kubernetes_cluster.rack]
-  value      = digitalocean_kubernetes_cluster.rack.name
+  value = digitalocean_kubernetes_cluster.rack.name
 }
 
 output "token" {
-  depends_on = [digitalocean_kubernetes_cluster.rack, null_resource.delay_token]
```
heronrs, line 22: I think we should keep the `null_resource.delay_token`

Twsouza:
CI: https://github.com/convox/convox/actions/runs/3394733837/jobs/5645791225

terraform/cluster/do/outputs.tf
Twsouza: my mistake

heronrs (approved):

PR #529, opened 2022-11-09
Feat-692: Add custom tags support for v3

nightfury1204:
### What is the feature/fix?
https://github.com/convox/issues-private/issues/692

NLB doesn't updates the tags: https://github.com/kubernetes/cloud-provider-aws/issues/113

https://user-images.githubusercontent.com/12232198/200738462-6baed797-0393-486a-b9fe-2fee2dcf9d23.mp4

no, just adding tag

### How to use/test it?
- update or install rack of this branch
- `convox rack params set tags="key1=val1,key2=val2"`

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [x] Documentation updated

### DIFF
--- a/pkg/cli/rack.go
+++ b/pkg/cli/rack.go
@@ -104,6 +104,16 @@ func validateParams(params map[string]string) error {
 		return errors.New("to schedule your rack to turn on/off you need both ScheduleRackScaleDown and ScheduleRackScaleUp parameters")
 	}
 
+	// format: "key1=val1,key2=val2"
+	if tags, has := params["tags"]; has {
+		tList := strings.Split(tags, ",")
+		for _, p := range tList {
+			if len(strings.Split(p, "=")) != 2 {
+				return errors.New("invalid value for tags param")
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/terraform/cluster/aws/main.tf
+++ b/terraform/cluster/aws/main.tf
@@ -57,6 +57,7 @@ resource "aws_eks_cluster" "cluster" {
 
   name     = var.name
   role_arn = aws_iam_role.cluster.arn
+  tags     = local.tags
   version  = var.k8s_version
 
   vpc_config {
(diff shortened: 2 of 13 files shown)


nightfury1204:
https://github.com/convox/convox/actions/runs/3421050108

pkg/cli/rack.go
```diff
@@ -104,6 +104,16 @@ func validateParams(params map[string]string) error {
 		return errors.New("to schedule your rack to turn on/off you need both ScheduleRackScaleDown and ScheduleRackScaleUp parameters")
 	}
 
+	// format: "key1=val1,key2=val2"
+	if tags, has := params["tags"]; has {
```
yuriadams, line 108: hey, nahid, do you think a validation like this it could be done by regex? I know, a lot of people are not fan of regex but it seems cleaner for me.

nightfury1204: But it's hard to understand what that regex is for by looking
do you have the regex expression for it?

heronrs: Yeah, if it's some obscure regexp I'd rather just keep it like this.
Even with regexp, I think we'd still need to do some type of check to compare the found groups with the actual string.

terraform/cluster/aws/main.tf
```diff
@@ -57,6 +57,7 @@ resource "aws_eks_cluster" "cluster" {
 
   name     = var.name
   role_arn = aws_iam_role.cluster.arn
+  tags     = local.tags
```
heronrs, line 60: I'm not sure if you already tried that, but seems like applying the tags to the AWS provider is cleaner(it will propagate to all created resources)
https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block

heronrs (commented):
Suggestion for maybe a cleaner solution

heronrs:
@nightfury1204 forgot to mention that not sure what happened but the video is all back screen

terraform/cluster/aws/main.tf
nightfury1204: @heronrs I checked it misses the ec2 tagging

heronrs: Should we explicitly declare `tags` in the `locals` block or how is it getting the values? Or should it be `var.tags`?

heronrs (commented):
Looks good! Just a code question

terraform/cluster/aws/main.tf
nightfury1204: It's in the another file https://github.com/convox/convox/blob/d2b22a9cceeceb80fa0a552fcb61ab075cdd6657/terraform/cluster/aws/network.tf#L3

since it were there before, so i didn't change it

heronrs (dismissed):

PR #530, opened 2022-11-09
Fix resource URL mapping

slaskis:
We use a product called metabase and previously we could deploy this nice little convox.yml:

```yml
resources:
    database:
      type: postgres

services:
  metabase:
    image: metabase/metabase:v0.44.6
    environment:
      - JAVA_TOOL_OPTIONS=-Xmx1g
    port: 3000
    resources:
      - database:MB_DB_CONNECTION_URI
```

But after #481 it will not work anymore because it changed that the exposed env var from a resource from `URL` to whatever is at the end of the env var we want to map it to, in our example `URI`. 

I don't think that was intentional as the "contract" for a resource seemed to be that it exposes a `URL` env var.

### DIFF
--- a/provider/k8s/process.go
+++ b/provider/k8s/process.go
@@ -428,14 +428,12 @@ func (p *Provider) podSpecFromService(app, service, release string) (*ac.PodSpec
 			c.Image = fmt.Sprintf("%s:%s.%s", repo, service, r.Build)
 
 			for _, r := range s.ResourceMap() {
-				parts := strings.Split(r.Env, "_")
-				key := parts[len(parts)-1]
 				c.Env = append(c.Env, ac.EnvVar{
 					Name: r.Env,
 					ValueFrom: &ac.EnvVarSource{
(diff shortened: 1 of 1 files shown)


heronrs:
@slaskis thanks for submitting the PR. 
@Twsouza will have a look at it to make sure it won't break the new functionality and meanwhile I'm merging it into a "pre" branch so I can have an rc for you to test. I will update you asap

heronrs:
@slaskis you can try with this one https://github.com/convox/convox/releases/tag/url-mapping-rc

Twsouza:
@slaskis thanks for pointing this out, I have a fix opened to your problem. Here's the PR in case you want to check it out: https://github.com/convox/convox/pull/533

Twsouza:
@slaskis the fix is out, see:

- https://github.com/convox/convox/releases/tag/3.6.10
- https://github.com/convox/convox/releases/tag/3.7.3
- https://github.com/convox/convox/releases/tag/3.8.2

PR #533, opened 2022-11-17
Fix resource custom link name

Twsouza:
### What is the feature/fix?

As described [here](https://github.com/convox/convox/pull/530), when you have a resource linked with a different name than ending with `URL` it breaks because doesn't find the correct key. 

In this case, it should use the default CM key instead of using the last resource link name.

image

No.

### How to use/test it?

- Install a rack with the fix
- Deploy an app using a custom resource link name (see convox.yml below)

```
resources:
  my-db:
    type: postgres
services:
  web:
    build: .
    port: 80
    resources:
      - my-db:CUSTOM_DATABASE
```

It should deploy with no errors.

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/manifest/service.go
+++ b/pkg/manifest/service.go
@@ -185,6 +185,19 @@ type ServiceResource struct {
 	Env  string
 }
 
+func (sr ServiceResource) GetConfigMapKey() string {
+	parts := strings.Split(sr.Env, "_")
+	key := parts[len(parts)-1]
+
+	for _, en := range AdditionalEnvNames {
+		if key == en {
+			return key
+		}
+	}
+
+	return DEFAULT_RESOURCE_ENV_NAME
+}
+
 func (s Service) AnnotationsMap() map[string]string {
 	annotations := map[string]string{}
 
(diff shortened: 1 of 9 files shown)


pkg/manifest/service.go
```diff
@@ -185,6 +185,19 @@ type ServiceResource struct {
...
+	parts := strings.Split(sr.Env, "_")
+	key := parts[len(parts)-1]
+
+	for _, en := range AdditionalEnvNames {
+		if key == en {
+			return key
+		}
+	}
+
+	return DEFAULT_RESOURCE_ENV_NAME
```
nightfury1204, line 198: Sorry can you give some context about it?
why if doesn't match with anything we are returning `DEFAULT_RESOURCE_ENV_NAME`?

Twsouza: sure thing!
when you create a resource, it will create a [ConfigMap](https://github.com/convox/convox/blob/0116fba7819c4284ba6103482adf83317e8eefa6/provider/k8s/template/resource/postgres.yml.tmpl#L12) with following keys: `URL`, `USER`, `PASS`, `HOST`, `PORT` and `NAME`.

The current behavior is to get the last part of the "linking name" and reference it in the environment variables pointing to the ConfigMap key. For example, setting `my-db:CUSTOM_DATABASE` in the service will create an environment variable called `CUSTOM_DATABASE` pointing to the `DATABASE` key in the ConfigMap. The database key doesn't exist in the CM and breaks the deployment.

To fix this, in case the user inserts a custom linking name, it will default to the URL key in the config map.

nightfury1204 (approved):
:fire:

yuriadams (approved):

heronrs (approved):

PR #534, opened 2022-11-18
feat-673: Manage eks addons

nightfury1204:
### What is the feature/fix?
https://github.com/convox/issues-private/issues/673

Rack update doesn't update the addons , so
Manage aws eks addons: vpc-cni, kube-proxy and coredns

**yes**
Once you upgrade to this version you can't downgrade, since addons are now managed by tf, so downgrading to lower version will delete those addons since they are not tf managed before and thus cluster breaks

### How to use/test it?
Install or upgrade your rack to this version

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/terraform/cluster/aws/variables.tf
+++ b/terraform/cluster/aws/variables.tf
@@ -11,7 +11,7 @@ variable "cidr" {
 }
 
 variable "coredns_version" {
-  type = string
+  type    = string
   default = null
 }
 
@@ -28,17 +28,17 @@ variable "internet_gateway_id" {
 }
 
 variable "key_pair_name" {
-  type = string
+  type    = string
   default = ""
 }
 
 variable "kube_proxy_version" {
-  type = string
+  type    = string
   default = null
 }
 
 variable "k8s_version" {
-  type = string
+  type    = string
   default = "1.21"
 }
 
@@ -71,6 +71,6 @@ variable "vpc_id" {
 }
 
 variable "vpc_cni_version" {
-  type = string
+  type    = string
   default = null
 }
--- a/terraform/cluster/aws/variables.tf
+++ b/terraform/cluster/aws/variables.tf
@@ -10,6 +10,11 @@ variable "cidr" {
   default = "10.1.0.0/16"
 }
 
+variable "coredns_version" {
+  type = string
+  default = null
+}
+
 variable "gpu_type" {
   default = false
 }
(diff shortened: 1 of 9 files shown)


terraform/system/aws/main.tf
```diff
@@ -60,6 +60,9 @@ module "cluster" {
   private             = var.private
   tags                = local.tag_map
   vpc_id              = var.vpc_id
+  vpc_cni_version = var.vpc_cni_version
```
yuriadams, line 63: just the `=`s are not aligned. The text editor should format this.

Twsouza (commented):
I have some thoughts 🤔 

What terraform CLI version you used to test it? Did you get any `error: cycle` while updating?
(I just want to check if the error cycle was solved)

Is it possible to update the add-on for GKE/DO/other cloud providers?

Thanks!

one nitpick, please format the terraform files.

nightfury1204:
> 



> I have some thoughts thinking
> 
> What terraform CLI version you used to test it? Did you get any `error: cycle` while updating? (I just want to check if the error cycle was solved)
> 
> Is it possible to update the add-on for GKE/DO/other cloud providers?
> 
> Thanks!
> 
> one nitpick, please format the terraform files.

@Twsouza
cli version locally in my pc `1.3.4`, but i also used the `convox rack update <version>` command. Also no i didn't get any cycle error

> Is it possible to update the add-on for GKE/DO/other cloud providers?

I didn't find anything explicit about that in terraform, probably need to look into that more and support it in another task

Twsouza:
thanks, @nightfury1204 
could you create a ticket to look at the add-on for the other cloud providers?

terraform/cluster/aws/variables.tf
```diff
@@ -10,6 +10,11 @@ variable "cidr" {
   default = "10.1.0.0/16"
 }
 
+variable "coredns_version" {
+  type = string
+  default = null
+}
```
Twsouza, line 16: this file as well

nightfury1204:
> thanks, @nightfury1204 could you create a ticket to look at the add-on for the other cloud providers?

done https://github.com/convox/issues-private/issues/714

Twsouza (approved):
LGTM

heronrs (approved):

PR #537, opened 2022-11-21
GPU reservation requirements

beastawakens:
### What is the feature/fix?

Allow services to reserve gpu resources.  Also handles bug within nvidia-docker with regards to cgroups where specifying cpu/memory limits/requests at the same time as gpu causes container crashes. 

No

### How to use/test it?

See test yaml / docs.

### Checklist
- [x] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [x] Documentation updated

### DIFF
--- a/docs/reference/primitives/app/service.md
+++ b/docs/reference/primitives/app/service.md
@@ -152,11 +152,22 @@ services:
 | Attribute | Type   | Default | Description                                                                                                   |
 | --------- | ------ | ------- | ------------------------------------------------------------------------------------------------------------- |
 | **count**   | number | 1       | The number of [Processes](/reference/primitives/app/process) to run for this Service. For autoscaling use a range, e.g. **1-5**        |
-| **cpu**     | number | 128     | The number of CPU units to reserve for [Processes](/reference/primitives/app/process) of this Service where 1024 units is a full CPU |
-| **memory**  | number | 256     | The number of MB of RAM to reserve for [Processes](/reference/primitives/app/process) of this Service                                |
+| **cpu**     | number | 256     | The number of CPU units to reserve for [Processes](/reference/primitives/app/process) of this Service where 1024 units is a full CPU |
+| **gpu**     | map    |         | The number/type of GPUs to reserve for [Processes](/reference/primitives/app/process) of this Service  |
+| **memory**  | number | 512     | The number of MB of RAM to reserve for [Processes](/reference/primitives/app/process) of this Service                                |
 | **targets** | map    |         | Target metrics to trigger autoscaling                                                                         |
 
 > Specifying **scale** as a number will set the **count** and leave the other values as defaults.
+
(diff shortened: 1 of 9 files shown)


beastawakens:
@heronrs @Twsouza @nightfury1204 @yuriadams thoughts? 😄

docs/reference/primitives/app/service.md
```diff
@@ -152,12 +152,23 @@ services:
...
 
+### scale.gpu
+
+| Attribute | Type   | Default | Description                                                                                |
+| --------- | ------ | ------- | ------------------------------------------------------------------------------------------ |
+| **count**  | number |        | The number of GPUs to reserve for [Processes](/reference/primitives/app/process) of this Service    |
+| **vendor** | string | nvidia | The GPU vendor to target for [Processes](/reference/primitives/app/process) of this Service |
+
+> Specifying **gpu** as a number will set the **count** and leave the vendor as default.
+> Specifying a **gpu** value and not specifying the cpu or memory to reserve will remove their defaults to purely reserve based on GPU.
```
Twsouza, line 170: I guess it only works with certain types of instances, should the documentation have a disclaimer about that?

Twsouza (commented):
E2E tests are running, I have one comment (below)

docs/reference/primitives/app/service.md
beastawakens: If you're not running on a GPU enabled instance then k8s would either ignore this, or just not be able to place your pod I guess.  I'll add a little note though to make it clear!

pkg/manifest/manifest.go
```diff
@@ -169,12 +169,18 @@ func (m *Manifest) ApplyDefaults() error {
 			m.Services[i].Scale.Count = ServiceScaleCount{Min: 1, Max: 1}
 		}
 
-		if m.Services[i].Scale.Cpu == 0 {
-			m.Services[i].Scale.Cpu = DefaultCpu
+		if m.Services[i].Scale.Gpu.Count == 0 {
```
heronrs, line 172: Just wondering why is that so. Couldn't we still set the default CPU and Memory values regardless? Or I guess that's a pattern when running GPU-specific pods?

heronrs (commented):
@beastawakens can you add a very simple gpu scale based app in test.sh? I know the instances prob won't support it but it's a good smoke test to detect that unsupported node instances won't break.

pkg/manifest/manifest.go
beastawakens: Yeah, there's a known issue/pattern where you should only specify GPU resources otherwise the nvidia-docker gets confused 🤷🏼‍♂️

heronrs (approved):
LGTM.
@beastawakens I know the Deepsource Go errors were not introduced by you, but if you find the time to clean them up it would be awesome :100:

PR #546, opened 2022-12-07
[Feat 729] Ability to Control LetsEncrypt Renewal Frequency

yuriadams:
### What is the feature/fix?
[Task 729](https://github.com/convox/issues-private/issues/729)

Ability to control LetsEncrypt renewal frequency.
There are two levels of control: 

1. The rack component, it will be done by terraform.
2. Services, it will be controlled by the convox.yaml file.

No, I had the concern to set the default values that is alread in use.

### How to use/test it?
Add param `cert_duration` for the terraform part.

In the convox.yaml, It was created a field `certDuration` that will control the renew for the service perspective.

### DIFF
--- a/provider/k8s/release.go
+++ b/provider/k8s/release.go
@@ -341,7 +341,8 @@ func (p *Provider) releaseTemplateIngress(a *structs.App, ss manifest.Services, 
 
 	items := [][]byte{}
 
-	for _, s := range ss {
+	for i := range ss {
+		s := ss[i]
 		ans, err := p.Engine.IngressAnnotations(s.Certificate.Duration)
 		if err != nil {
 			return nil, errors.WithStack(err)
--- a/provider/k8s/release.go
+++ b/provider/k8s/release.go
@@ -334,11 +334,6 @@ func (p *Provider) releaseTemplateCA(a *structs.App, ca *v1.Secret) ([]byte, err
 }
 
 func (p *Provider) releaseTemplateIngress(a *structs.App, ss manifest.Services, opts structs.ReleasePromoteOptions) ([]byte, error) {
-	ans, err := p.Engine.IngressAnnotations(a.Name)
-	if err != nil {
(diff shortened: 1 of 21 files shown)


beastawakens:
I would urge the convox.yml configuration to just be `certificate` or similar to maintain the pattern with all the other service parameters, without camel-casing it. You may want to expand this section later on with sub-fields. You could then specify `duration` as a specific sub-field if you want. 
I've been thinking about wildcard domains recently so this could be useful 😏

provider/k8s/release.go
```diff
@@ -334,19 +334,20 @@ func (p *Provider) releaseTemplateCA(a *structs.App, ca *v1.Secret) ([]byte, err
...
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 
 	items := [][]byte{}
 
-	for _, s := range ss {
+	for i := range ss {
+		s := ss[i]
+		ans, err := p.Engine.IngressAnnotations(s.Certificate.Duration)
```
nightfury1204, line 346: I can see you defined rack param and also service wise parameter for cert renewal.
For service, if it is not defined but rack cert param is defined, then I think we should use it as the default value.

yuriadams: hey @nightfury1204 I thought the rack value is used by default for services.

nightfury1204: not sure can you check please?

yuriadams: yeah, I checked, I was just double checking I you've seen anything that I haven't.

nightfury1204: don't you think we should use rack cert renewal period as default?

yuriadams: The default is the rack renewal period. If there's nothing setup for service, it uses the default rack period.

nightfury1204 (approved):

PR #549, opened 2023-01-03
remove hostNetwork config for agents

yuriadams:
### What is the feature/fix?

[task 804] (https://github.com/convox/issues-private/issues/804)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 # these owners will be the default owners for approving PRs to merge into the `master` branch
 
-*       @master-approvers
+*       @convox/master-approvers
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# these owners will be the default owners for approving PRs to merge into the `master` branch
+
+*       @master-approvers
--- a/provider/k8s/template/app/service.yml.tmpl
+++ b/provider/k8s/template/app/service.yml.tmpl
@@ -85,9 +85,6 @@ spec:
           - "svc.cluster.local"
           - "cluster.local"
       {{ end }}
-      {{ if .Service.Agent.Enabled }}
-      hostNetwork: true
-      {{ end }}
       serviceAccountName: {{.Service.Name}}
       shareProcessNamespace: {{.Service.Init}}
       terminationGracePeriodSeconds: {{$.Service.Termination.Grace}}
--- a/docs/configuration/agents.md
+++ b/docs/configuration/agents.md
@@ -40,7 +40,9 @@ Each [Process](/reference/primitives/app/process) will have the IP address of it
 
 In the example above, any [Process](/reference/primitives/app/service) on the same Rack can communicate
 with the `telemetry` Agent running on its [Instance](/reference/primitives/rack/instance) using the
-following endpoints:
+following endpoints**:
 
 * `udp://$INSTANCE_IP:8125`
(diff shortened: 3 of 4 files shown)


provider/k8s/template/app/service.yml.tmpl
```diff
@@ -85,9 +85,6 @@ spec:
           - "svc.cluster.local"
           - "cluster.local"
       {{ end }}
-      {{ if .Service.Agent.Enabled }}
```
heronrs, line 88: Did you test agent [communication](https://docs.convox.com/configuration/agents/) after this change?
If it doesn't work we need to provide another solution. Probably having a k8s service for the agent like we do for other apps(if it doesn't create one already)

yuriadams: hey @heronrs, I'm still making manual testing this. I created a draft PR to have a look at the checks down here. 
On the first tests, there was no change in the communication between pods in the same rack, but I'm not confident 100%.

Twsouza:
@yuriadams can you change the base branch to `staging`? thanks

provider/k8s/template/app/service.yml.tmpl
heronrs: Hey,
> It worked, another processes running in another instances are not visible anymore.

What do you mean by that? Were such processes visible before the change?
I'm not sure about the svc because routing will go round robin way and by using the instance IP you guarantee the service will always reach the agent running in the same node.

yuriadams: there was one case that I could communicate: 

-  I had an deployed app, I applied the change to remove the hostNetwork from the agent, and I still could communicate. Even I could make a request for an agent running at the same node.
-  Now, I deployed another app with this change already applied, this way I could not communicate by IP with another nodes in the same rack.

heronrs: I think being able to communicate with the same node is fine - we should document that in the release notes

As for the svc is not much about security but the fact that workloads on the same node could be reaching an agent in a different node - you could have agents in one node receiving a lot of traffic while others could be idle.

yuriadams: got it. I'll update the docs as well and include here, on this PR.

.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# these owners will be the default owners for approving PRs to merge into the `master` branch
+
+*       @master-approvers
```
heronrs, line 3: this seems to be wrong according to GH

heronrs (approved):
LGTM.
Let's add some notes in the release about the breaking change and that's a security improvement.
Deepsource is also failing

.github/CODEOWNERS
yuriadams: I merged with master before change the branch base. I'll revert this part, thanks!

yuriadams:
> LGTM. Let's add some notes in the release about the breaking change and that's a security improvement. Deepsource is also failing

The deepsource, heron, it's scanning all project, not only the recent changes. So, really weird this behaviour. And also, almost impossible to fix the report. For example:

Screenshot 2023-01-04 at 13 36 20

.github/CODEOWNERS
yuriadams: @Twsouza, I have just checked this file and it's the same as master. Can I ignore this warning?

heronrs (approved):

PR #550, opened 2023-01-04
Update registry docs

heronrs:
I just tested this and looks like the `https://` suffix is not added automatically.

I added an internal ticket to fix it


Twsouza:
@heronrs won't this break `registries delete`?
@ntner added this 2 months ago

Twsouza (approved):

PR #560, opened 2023-01-31
3.10.4

Twsouza:
### What is include in the release?

- [Fix 729] Fix name of the certificate property on convox.yaml
- OIDC providers (CI)
- Show cli expiration error


ntner (approved):
Tested:
Multiple variations of parameters for install and upgrade to this version
All parameter updates successful 
New application install and working at version
Existing application working after upgrade
Clean upgrade from 3.6.X
Review and Deploy Workflows working at version and across update

PR #563, opened 2023-02-03
Create rack using existing subnets

lucasmacedot:
### What is the feature/fix?

Currently there's no way to use existing subnets to create the rack.
This PR is to add this option to the CLI, so the user can pass existing private and public subnets, and convox creates the rack using these subnets.

The subnets are passed as a string, and the main module (`system/aws`) splits the string into an array of string, then pass this array to subsequent modules.

This approach was chosen because I didn't want to change the golang part of the cli. I would have to change the vars from `map[string]string` to `map[string]interface{}`, and this would then require change not only the terraform part, but also the console and the direct part, which would then require to change the web API.

** Describe the bug and the solution. **
** Any additional information **

I don't think there is a breaking change

### How to use/test it?

** Describe how to test or use the feature **
The goal is to run `convox rack install aws convox-lucasmacedo vpc_id=vpc-1234567890 private_subnets_ids=subnet-02225d6c02ebfcb31,subnet-068ebb46c7b2f6bdc,subnet-075b53ddf4910129c public_subnets_ids=subnet-0e7b000faba6a77f4,subnet-01518d03c99ea3cb2,subnet-0e3b2be5cc35253ca internet_gateway_id=igw-0c858f953bcf1798d`

And the module created on terraform then be:

```terraform
module "system" {
  source              = "github.com/convox/convox//terraform/system/aws?ref=3.10.3"
  name                = "convox-lucasmacedo"
  release             = "3.10.3"
  vpc_id              = "vpc-1234567890"
  private_subnets_ids = "subnet-02225d6c02ebfcb31,subnet-068ebb46c7b2f6bdc,subnet-075b53ddf4910129c"
  public_subnets_ids  = "subnet-0e7b000faba6a77f4,subnet-01518d03c99ea3cb2,subnet-0e3b2be5cc35253ca"
  internet_gateway_id = "igw-0c858f953bcf1798d"
}
```

### Caveats
- It is assumed that the route table, internet gateway, nat gateway ids are already setup, so in the PR it is used data to get them. The gateway with the "0.0.0.0/0" route to public internet, etc.
- The internet gateway id needs to be passed, because the current data from terraform can't get internet gateway ids from just the associated subnets

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated


terraform/cluster/aws/main.tf
```diff
@@ -102,7 +102,7 @@ resource "aws_eks_node_group" "cluster" {
   instance_types  = split(",", random_id.node_group.keepers.node_type)
   node_group_name = "${var.name}-${local.availability_zones[count.index]}-${count.index}${random_id.node_group.hex}"
   node_role_arn   = random_id.node_group.keepers.role_arn
-  subnet_ids      = [var.private ? aws_subnet.private[count.index].id : aws_subnet.public[count.index].id]
+  subnet_ids      = [var.private ? var.private_subnets_ids[count.index] : aws_subnet.public[count.index].id]
```
Renatovnctavares, line 105: I think will not be able to use the default subnet created by this module  anymore `aws_subnet.private[count.index].id` 
 can we use `local.private_subnets_ids[count.index]` instead and have this logic there?
 or add a double condition here like aws_subnet.private resource?

terraform/cluster/aws/network.tf
```diff
@@ -4,7 +4,8 @@ locals {
     Name = var.name
     Rack = var.name
   })
-  vpc_id = var.vpc_id == "" ? aws_vpc.nodes[0].id : var.vpc_id
+  vpc_id              = var.vpc_id == "" ? aws_vpc.nodes[0].id : var.vpc_id
+  private_subnets_ids = length(var.private_subnets_ids) != 0 ? var.private_subnets_ids : aws_subnet.private[*].id
```
Renatovnctavares, line 8: wondering what would be the value of `local.private_subnets_ids` in case of `var.private = false`. and `var.private_subnets_ids` is empty, 
null?   

also, I did not find where we are using `local.private_subnets_ids`

terraform/cluster/aws/outputs.tf
```diff
@@ -34,7 +34,7 @@ output "route_table_public" {
 }
 
 output "subnets" {
-  value = aws_subnet.private.*.id
+  value = length(var.private_subnets_ids) == 0 ? aws_subnet.private.*.id : var.private_subnets_ids
```
Renatovnctavares, line 37: LGTM  but 
in case var.private=false the main resource `aws_eks_node_group.cluster` will use a public subnet. but it is not defined in the output.

terraform/cluster/aws/variables.tf
```diff
@@ -84,3 +84,13 @@ variable "vpc_cni_version" {
   type    = string
   default = null
 }
+
+variable "private_subnets_ids" {
+  type    = list(string)
+  default = []
+}
+
+variable "public_subnets_ids" {
```
Renatovnctavares, line 93: public_subnets_ids is not being used

terraform/system/aws/variables.tf
```diff
@@ -123,6 +123,16 @@ variable "vpc_cni_version" {
   default = "v1.11.4-eksbuild.1"
 }
 
+variable "private_subnets_ids" {
+  type    = list(string)
+  default = []
+}
+
+variable "public_subnets_ids" {
```
Renatovnctavares, line 131: public_subnets_ids is not being used

Renatovnctavares (changes requested):

terraform/cluster/aws/outputs.tf
lucasmacedot: There's no need to define it on the output, since it is on the same module, the `aws_eks_node_group.cluster` can reference the resource directly, without using the output.

I don't know where this output is used, but originally it was referencing the private subnets, so I don't think there's a need to change it to use public subnets

Twsouza:
hi @lucasmacedot! thanks for your contribution!
could you change the base branch to `convox:staging`?

Twsouza:
@lucasmacedot I was able to change the base branch, we will run some tests and review your PR soon.

pkg/cli/login.go
```diff
@@ -50,9 +49,6 @@ func Login(rack sdk.Interface, c *stdcli.Context) error {
 
 	id, err := cl.Auth()
 	if err != nil {
-		if strings.Contains(err.Error(), "cli token is expired") {
-			return err
-		}
```
Twsouza, line 55: please keep the message

pkg/rack/console.go
```diff
@@ -388,10 +388,6 @@ func listConsole(c *stdcli.Context) ([]Console, error) {
 	}
 
 	rs, err := cc.RackList()
-	if err != nil && strings.Contains(err.Error(), "cli token is expired") {
-		return nil, err
-	}
```
Twsouza, line 393: please keep the message

terraform/cluster/aws/network.tf
```diff
@@ -1,10 +1,17 @@
...
-  internet_gateway_id = var.internet_gateway_id == "" ? aws_internet_gateway.nodes[0].id : var.internet_gateway_id
   tags = merge(var.tags, {
     Name = var.name
     Rack = var.name
   })
   vpc_id = var.vpc_id == "" ? aws_vpc.nodes[0].id : var.vpc_id
+
+  private_subnets_ids  = length(var.private_subnets_ids) == 0 ? aws_subnet.private[*].id : var.private_subnets_ids
+  private_route_tables = length(var.private_subnets_ids) == 0 ? aws_route_table.private[*].id : data.aws_route_table.private[*].id
+  #nat_gateway_ids       = length(var.private_subnets_ids) == 0 ? aws_nat_gateway.private[*].id : data.aws_nat_gateway.private[*].id
```
Twsouza, line 10: delete if is not in use

Twsouza (approved):
lgtm

Twsouza (changes requested):
see my previous review

lucasmacedot:
> see my previous review

Solved

terraform/cluster/aws/network.tf
```diff
@@ -144,18 +160,24 @@ resource "aws_eip" "nat" {
...
 }
 
+data "aws_route_table" "private" {
+  count     = length(var.private_subnets_ids)
+  subnet_id = var.private_subnets_ids[count.index]
+}
+
 resource "aws_route_table" "private" {
-  count = var.private ? local.network_resource_count : 0
+  #count = var.private ? local.network_resource_count : 0
```
Twsouza, line 179: nitpick: could you delete this comment?

lucasmacedot: Solved

PR #579, opened 2023-02-21
Skip elastic GPU tagging eu-west-3

yuriadams:
### What is the feature/fix?

[Task 893](https://github.com/convox/issues-private/issues/893)

No


ntner (approved):
looks good to me

PR #589, opened 2023-03-03
Change rack name

Twsouza:
### What is the feature/fix?

Enables change on rack names using CLI.

No, new feature.

### How to use/test it?

1. Install the RC CLI
2. Install rack using the CLI: `convox rack install rackname -v RC` (RC is being created)
3. Move the rack `convox rack mv rackname newrackname`
4. Switch to the new name: `convox switch newrackname`
5. Run `convox rack` and the name should be `newrackname`.

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/k8s/system.go
+++ b/provider/k8s/system.go
@@ -15,6 +15,7 @@ import (
 
 func (p *Provider) SystemGet() (*structs.System, error) {
 	status := "running"
+	fmt.Printf("provider SystemGet: %+v\n", p)
 
 	// status, err := p.Engine.SystemStatus()
 	// if err != nil {
@@ -35,7 +36,7 @@ func (p *Provider) SystemGet() (*structs.System, error) {
 
 	s := &structs.System{
 		Domain:   fmt.Sprintf("router.%s", p.Domain),
-		Name:     p.Name,
+		Name:     p.RackName,
 		Provider: p.Provider,
 		Status:   status,
 		Version:  p.Version,
--- a/provider/k8s/system.go
+++ b/provider/k8s/system.go
@@ -15,7 +15,6 @@ import (
 
 func (p *Provider) SystemGet() (*structs.System, error) {
 	status := "running"
-	fmt.Printf("provider SystemGet: %+v\n", p)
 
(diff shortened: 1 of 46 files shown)


terraform/api/aws/variables.tf
```diff
@@ -32,6 +32,10 @@ variable "name" {
   type = string
 }
 
+variable "rack_name" {
+  type = string
+}
```
Twsouza, line 37: Since name is used by the k8s cluster and can't be changed, this variable will be responsible to have the current rack name.

Twsouza:
I found some issues while moving the rack in the console, looking at it.

provider/k8s/system.go
```diff
@@ -35,7 +36,7 @@ func (p *Provider) SystemGet() (*structs.System, error) {
 
 	s := &structs.System{
 		Domain:   fmt.Sprintf("router.%s", p.Domain),
-		Name:     p.Name,
+		Name:     p.RackName,
```
nightfury1204, line 39: From the changes i can understand. This var is set only when you ran the `mv` command.
will this have proper value for freshly installed rack of this version?

nightfury1204 (changes requested):

provider/k8s/system.go
Twsouza: it will also be set on rack install, the problem with the [rack variables](https://github.com/convox/convox/blob/bd16c3cac73cb4bd264db88809fa9eeb8115e558/terraform/system/aws/variables.tf#L80) is that we can't change or all resources will be destroyed, so the RackName will be the actual rack name and name is the "original" name.

nightfury1204: little bit confused about the rack name and original name?

nightfury1204:
@Twsouza it's ready?

Twsouza:
> @Twsouza it's ready?

it's failing on DO and GCP, I'll have a look later.

nightfury1204 (approved):

yuriadams (approved):

PR #591, opened 2023-03-09
IMDSv1 to IMDSv2 on v3 racks

yuriadams:
### What is the feature/fix?

Add field on eks cluster to control if Whether or not the metadata service requires session tokens, also referred to as Instance Metadata Service Version 2 (IMDSv2). Can be optional or required.

Task Link in [Asana](https://app.asana.com/0/1203637156732418/1204066841595405/f)
 
No

### How to use/test it?

There's a new parameter available on rack installation called: imds_http_tokens that can be `optional` or `required`.

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/terraform/cluster/aws/main.tf
+++ b/terraform/cluster/aws/main.tf
@@ -96,15 +96,15 @@ resource "aws_eks_node_group" "cluster" {
 
   count = var.high_availability ? 3 : 1
 
-  ami_type        = var.gpu_type ? "AL2_x86_64_GPU" : var.arm_type ? "AL2_ARM_64" : "AL2_x86_64"
-  capacity_type   = var.node_capacity_type
-  cluster_name    = aws_eks_cluster.cluster.name
-  instance_types  = split(",", random_id.node_group.keepers.node_type)
-  node_group_name = "${var.name}-${local.availability_zones[count.index]}-${count.index}${random_id.node_group.hex}"
-  node_role_arn   = random_id.node_group.keepers.role_arn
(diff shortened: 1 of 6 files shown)


nightfury1204 (dismissed):

terraform/system/aws/main.tf
```diff
@@ -55,6 +55,7 @@ module "cluster" {
   key_pair_name            = var.key_pair_name
   kube_proxy_version       = var.kube_proxy_version
   k8s_version              = var.k8s_version
+  imds_http_tokens         = var.imds_http_tokens
```
nightfury1204, line 58: did we already has this support?

nightfury1204: sorry i don't see where it is being applied 
i only see these changes: https://github.com/convox/convox/pull/591/files

yuriadams: yes, it's true. Some files are missing. Thank for letting me know! :)

yuriadams: this is a rack parameter, so, the only change is in the tf files, right?

terraform/cluster/aws/main.tf
```diff
@@ -155,6 +155,10 @@ resource "aws_launch_template" "cluster" {
     }
   }
 
+  metadata_options{
```
nightfury1204, line 158: can you run `terraform fmt`?

nightfury1204 (approved):

PR #592, opened 2023-03-09
Replicates a v2 paradigm where you can have a minimum number of on-demand instances for guaranteed service availability but utilise Spot for most of your scaling needs.

beastawakens:
### What is the feature/fix?

v2 Racks had a great feature which allowed you to specify a minimum number of on-demand instances to guarantee service availability, but utilise spot instances for cost savings when your Rack scaled.  This tries to replicate that paradigm in v3 by introducing a `mixed` capacity type option which would define on-demand instances in the first node group, but use spots in the other two.
Also introduces two other parameters to control the minimum and maximum counts of those on-demand instances.

Example: set both `min_on_demand_count` and `min_on_demand_count` to 3 to always have exactly 3 instances running for core services.  2 other spot instances would initially be spun up as well, and those node groups would scale as desired.  If all the spot instances were taken offline due to spot market fluctuations, the 3 on-demand instances would always be running.

No.

### How to use/test it?

Set `node_capacity_type` to `mixed`.  See one node group created with on-demand instances, the other two with spots.
Set `min_on_demand_count` and `max_on_demand_count` to appopriate values to control the sizing of the on-demand node group.

### Checklist
- [n/a] New coverage tests
- [x] Unit tests passing
- [ ] E2E tests passing
- [x] Documentation updated


Twsouza (approved):

nightfury1204:
rebased it with staging

Twsouza (approved):

nightfury1204 (approved):

PR #598, opened 2023-03-23
Use SSO to create CLI token

yuriadams:
### What is the feature/fix?

Docs:

## configure

Configure necessary fields to make the authentication using `login` command.
Fields: 
- Provider: Which identity provider will be used to authenticate. Supported providers: `okta`.
- Client ID: The OpenID Connect client ID provided by your Identity Provider.
- Client secret: The OpenID Connect client secret provided by your Identity Provider.
- Issuer: A unique string that identifies the provider issuing a request.
- Callback URL: A unique string that identifies the callback url that the idp will redirect the request after authorize the customer.

### Usage
```html
    convox sso configure
```

or

```html
    convox sso configure -p PROVIDER -c CLIENT_ID -s CLIENT_SECRET -i ISSUER -u CALLBACK_URL
```

or we can add following environment variables:

```
SSO_PROVIDER
SSO_CLIENT_ID
SSO_CLIENT_SECRET
SSO_ISSUER
SSO_CALLBACK_URL
```

### Examples
```html
    convox sso configure
    SSO Provider: 
    SSO Client ID: 
    SSO Client Secret: 
    SSO Issuer:
    SSO Callback URL: 
    OK
```

## login

Authenticate your CLI using a identity provider.

### Usage
```html
    convox sso login
```

You will be redirected to the browser to your fill the credentials in your identity provider.

### Requirements
Is necessary to configure the identity provider to add the `convoxID` that refers with the user id in the console. Also, is required to add this field in the token claim. 
Also, configure the callback endpoint in the customer application, for example, `http://localhost:8090/authorization-code/callback`. Port and the path can be customized by the customer.

No

### Video Record

https://convox602-my.sharepoint.com/:u:/r/personal/yuri_adams_convox_com/Documents/Convox%20-%20SSO%20Demo.zip?csf=1&web=1&e=Qqro5W

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/cli/sso_test.go
+++ b/pkg/cli/sso_test.go
@@ -67,10 +67,6 @@ func (m *MockSsoProvider) ExchangeCode(r *http.Request, code string) structs.Sso
 	}
 }
 
-func (m *MockSsoProvider) Name() string {
-	return "name"
-}
-
 func (m *MockSsoProvider) Opts() structs.SsoProviderOptions {
 	return structs.SsoProviderOptions{
 		State: "state123",
--- a/pkg/structs/sso.go
+++ b/pkg/structs/sso.go
@@ -4,7 +4,6 @@ import "net/http"
 
 type SsoProvider interface {
 	ExchangeCode(r *http.Request, code string) SsoExchangeCode
-	Name() string
 	Opts() SsoProviderOptions
 	RedirectPath() string
 	VerifyToken(t string) error
--- a/sso/okta/okta.go
+++ b/sso/okta/okta.go
@@ -5,7 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/url"
 
@@ -19,10 +19,6 @@ type Okta struct {
 
 func Initialize(opts structs.SsoProviderOptions) (structs.SsoProvider, error) {
 	return &Okta{opts}, nil
-}
-
-func (o *Okta) Name() string {
-	return "okta"
 }
 
 func (o *Okta) Opts() structs.SsoProviderOptions {
@@ -63,7 +59,8 @@ func (o *Okta) ExchangeCode(r *http.Request, code string) structs.SsoExchangeCod
 
 	client := &http.Client{}
 	resp, _ := client.Do(req)
-	body, _ := ioutil.ReadAll(resp.Body)
+	body, _ := io.ReadAll(resp.Body)
+
 	defer resp.Body.Close()
 	var exchange structs.SsoExchangeCode
 	json.Unmarshal(body, &exchange)


nightfury1204 (changes requested):
@yuriadams nice job.

but userid as convoxID poses a security issue. it should't be a user id. since we once someone knows the user id, they can easily login in. also its hard to change if it gets compromised since it's act as an identifier of the user in our system and all the data associated with it.

also can every user in the sso okta can configure identity provider?

yuriadams:
> @yuriadams nice job.
> 
> but userid as convoxID poses a security issue. it should't be a user id. since we once someone knows the user id, they can easily login in. also its hard to change if it gets compromised since it's act as an identifier of the user in our system and all the data associated with it.
> 
> also can every user in the sso okta can configure identity provider?

@nightfury1204 thank you very much for your feedback.

First question:

This is a great point that I don't have the answer what's the the best way. A improvement I did this morning was add the convoxID as a token claim. So, this token will be sent to the console and the console will be responsable to check if the id is there, so the cli won't "know" the convox id because it will be in the token payload. So, your concern is more the person who's configuring the okta knows the user id? 

Second: 
No, actually the customer needs to configure his own okta application, we don't control. In the video record I just show which requirements the customer needs to do before configure our CLI.

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+		return err
+	}
+
+	if err := c.SettingWrite("client_secret", clientSecret); err != nil {
+		return err
+	}
+
+	if err := c.SettingWrite("issuer", issuer); err != nil {
+		return err
+	}
```
Twsouza, line 98: is it better to write all information into a single file?

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+
+	clientSecret, err := c.SettingRead("client_secret")
+	if err != nil {
+		return err
+	}
+
+	issuer, err := c.SettingRead("issuer")
+	if err != nil {
+		return err
+	}
```
Twsouza, line 122: it's creating multiple files inside of `~/.config/convox`

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+
+	server := &http.Server{Addr: "/authorization-code/callback"}
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		authCodeCallbackHandler(w, r, server, c, p)
+	})
+
+	port := fmt.Sprintf(":%s", "8090")
```
Twsouza, line 145: what if port 8090 is in use? maybe set this as an env var?

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+		os.Exit(1)
+	}
+
+	log.Println("Waiting for login... ")
+
+	time.Sleep(1 * time.Second)
+	sso.Openbrowser(p.RedirectPath())
+	time.Sleep(1 * time.Second)
+
+	server.Serve(l)
```
Twsouza, line 158: what happens if the user don't finish the process? will the server be open? is it possible to add a timeout option? after X min turn it down

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+	if issuer == "" {
+		c.Writef("SSO ISSUER: ")
+
+		issuer, err = c.ReadSecret()
+		if err != nil {
+			return err
+		}
+
+		c.Writef("\n")
+	}
```
Twsouza, line 82: the option to set the environment variables is not documented, can you add it plz?

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+	verificationError := p.VerifyToken(exchange.IdToken)
+
+	if verificationError != nil {
+		fmt.Println(verificationError)
+	}
+
+	if err := c.SettingWrite("bearer_token", exchange.AccessToken); err != nil {
+		fmt.Fprintf(w, "Could not set Bearer Token on the CLI settings")
+		return
+	}
```
Twsouza, line 190: use only one file to keep the sso data

pkg/cli/sso.go
```diff
@@ -0,0 +1,216 @@
...
+
+	w.Header().Set("Content-Type", "text/html")
+	w.Write(openSuccessHTMLFile())
+
+	cleanup(server)
+}
+
+func openSuccessHTMLFile() []byte {
+	file, err := os.Open("../../sso/templates/success.html")
```
Twsouza, line 199: this file won't be available in the user machine when they install the CLI, by default go doesn't include static files into the binary executable
to use the file you have to "pack" it inside the binary, we already have https://github.com/gobuffalo/packr in the project, you can use it to bundle the file and to be available in the CLI

Twsouza (changes requested):

pkg/cli/sso.go
```diff
@@ -0,0 +1,256 @@
...
+	done := make(chan bool)
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		authCodeCallbackHandler(w, r, c, p, done)
+	})
+
+	go func() {
+		l, err := net.Listen("tcp", fmt.Sprintf(":%s", port))
+		if err != nil {
+			fmt.Printf("snap: can't listen to port %s: %s\n", port, err)
```
nightfury1204, line 182: should it handle the error?

nightfury1204 (approved):

pkg/cli/sso.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	verificationError := p.VerifyToken(exchange.IdToken)
+
+	if verificationError != nil {
+		fmt.Println(verificationError)
+	}
+
+	if err := c.SettingWriteKey("sso", "bearer_token", exchange.AccessToken); err != nil {
+		fmt.Fprintf(w, "Could not set Bearer Token on the CLI settings")
```
Twsouza, line 240: What the user should do when this happens? Try to add more context here.

pkg/cli/sso.go
```diff
@@ -0,0 +1,264 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "text/html")
+	file, err := openAuthHTMLFile()
+	if err != nil {
+		fmt.Fprintf(w, "Could not open the success authentication html page")
```
Twsouza, line 247: What the user should do when this happens? Try to add more context here.

Twsouza (dismissed):

nightfury1204 (dismissed):

Twsouza:
@yuriadams is this good to review?

PR #608, opened 2023-04-05
Store rack install & rack install update telemetry data

yuriadams:
### What is the feature/fix?

[Task ticket](https://app.asana.com/0/1203637156732418/1204077860485348/f)

Screenshot 2023-03-31 at 14 10 27

Demo: https://convox602-my.sharepoint.com/:v:/g/personal/yuri_adams_convox_com/EQv78h8kYzxOs-Mv-6ySrVsBJH5G0ue-thrSH3G80dngng?e=9P8Pu8

Doc:  https://convox602-my.sharepoint.com/:w:/g/personal/yuri_adams_convox_com/ERKCC87MHOxLi6_3F8PqWqIBOWecg2nEg06tIbl9NA6ujQ?e=8bW5AX

No

### Usage
```html
    convox telemetry set true
```

### Examples
```html
    convox telemetry set true
    OK

    convox telemetry set false
    OK

    convox telemetry set blah
    ERROR: command accepts just 'true' or 'false' as argument
```

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated


provider/k8s/telemetry.go
```diff
@@ -0,0 +1,137 @@
...
+	trs, err := p.Cluster.CoreV1().ConfigMaps(p.Namespace).Get(context.TODO(), "telemetry-rack-sync", am.GetOptions{})
+	if err != nil {
+		fmt.Printf("could not find params sync configmap: %v: Creating configmap...", err)
+
+		var hadErrorOnCreate bool
+		var noParamsToSync map[string]interface{}
+		trs, hadErrorOnCreate, noParamsToSync = createNewSyncConfigMap(trp, trs, err, p)
+		if hadErrorOnCreate {
+			return noParamsToSync
+		}
```
Twsouza, line 45: if you still want to use `trs` I suggest to use a pointer, so in the `createNewSyncConfigMap` you can assigned directly to it.
also, `noParamsToSync` is always returned as nil, consider removing it.

provider/k8s/telemetry.go
```diff
@@ -0,0 +1,137 @@
...
+			} else {
+				toSync[s] = val
+			}
+		}
+	}
+
+	return toSync
+}
+
+func createNewSyncConfigMap(trp *v1.ConfigMap, trs *v1.ConfigMap, err error, p *Provider) (*v1.ConfigMap, bool, map[string]interface{}) {
```
Twsouza, line 90: consider removing `err error`, is not being used. You can create a new var inside of the method and not passing it.
the last parameter is always returned as nil, maybe remove it?

Twsouza: imho: `createNewSyncConfigMap` should receive only the provider, the trp and the trs and return only an error
inside of the method you can assing the new trs to the old trs, as both variables are pointers, this will work fine.

```suggestion
func createNewSyncConfigMap(p *Provider, trp *v1.ConfigMap, trs *v1.ConfigMap) (error) {
```

provider/k8s/telemetry_test.go
```diff
@@ -0,0 +1,194 @@
+package k8s_test
```
Twsouza, line 1: good one!

terraform/rack/k8s/main.tf
```diff
@@ -1,3 +1,8 @@
+locals {
+  config_data = jsondecode(file(var.telemetry_file))
+  send_telemetry = contains(keys(local.config_data), "telemetry") ? index(local.config_data["telemetry"], 0) != "false" : true
```
Twsouza, line 3: isn't it better just to check if `var.telemetry` is true or false?

Twsouza (changes requested):

pkg/rack/terraform.go
```diff
@@ -824,7 +825,8 @@ func terraformWriteTemplate(filename, version string, params map[string]interfac
 	t, err := template.New("main").Funcs(terraformTemplateHelpers()).Parse(`
 		module "system" {
 			source = "{{.Source}}"
-
+			settings = "{{.Settings}}"
```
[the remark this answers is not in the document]
yuriadams: to identify where's the vars.json file. Just the cli itself knows where this file is created. So, I have to pass as a variable ```
vars_file = "${var.settings}/vars.json"
```

File: terraform/rack/k8s/main.tf [hunk 6]
```diff
@@ -1,3 +1,8 @@
+locals {
+  config_data = jsondecode(file(var.telemetry_file))
```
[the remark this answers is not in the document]
yuriadams: thanks, @Twsouza, do we have an example for that?

terraform/system/do/variables.tf
```diff
@@ -60,10 +60,18 @@ variable "secret_key" {
...
+variable "settings" {
+  default = ""
+}
+
 variable "syslog" {
   default = ""
 }
 
+variable "telemetry" {
+  default = ""
```
Twsouza, line 72: set default to true

Twsouza (changes requested):

terraform/rack/gcp/variables.tf
```diff
@@ -55,6 +55,15 @@ variable "syslog" {
   default = ""
 }
 
+variable "telemetry" {
+  default = ""
+}
```
Twsouza, line 60: you can use this variable as boolean
https://www.devopsschool.com/blog/terraform-variable-boolean-type-explained/

terraform/rack/k8s/main.tf
```diff
@@ -47,3 +47,14 @@ DOCKER
 
   type = "kubernetes.io/dockerconfigjson"
 }
+
+resource "kubernetes_config_map" "telemetry_configuration" {
+  count = var.telemetry != "false" ? 1 : 0
```
Twsouza, line 52: use it as boolean

```suggestion
  count = var.telemetry ? 1 : 0
```

terraform/system/aws/main.tf
```diff
@@ -130,6 +130,8 @@ module "rack" {
   ssl_protocols       = var.ssl_protocols
   subnets             = module.cluster.subnets
   tags                = local.tag_map
+  telemetry           = var.telemetry
+  telemetry_file      = "${var.settings}/vars.json"
```
Twsouza, line 134: isn't it better if we call `telemetry_file` as `variables_file` or `settings_file`?

Twsouza (changes requested):
just a few more adjustments then we are ready to go

nightfury1204 (approved):

PR #617, opened 2023-04-14
Docs remove unused dns

Twsouza:
### What is the feature/fix?

Update documentation about DNS suffixes. The pattern `http://SERVICE:PORT` or `http://SERVICE.APP:PORT` is not in use.

N/A

### How to use/test it?

N/A

### Checklist

N/A

### DIFF
--- a/docs/configuration/service-discovery.md
+++ b/docs/configuration/service-discovery.md
@@ -62,9 +62,3 @@ The `web` [Service](/reference/primitives/app/service) could reach the `auth` [S
 > Note that the internal port of the `auth` [Service](/reference/primitives/app/service) is not receiving
 > automatic SSL termination. If you want this connection to be encrypted you would need to handle SSL
 > inside the [Service](/reference/primitives/app/service).
-
-DNS search suffixes are automatically configured for internal hostnames on a Rack. The following URLs would
-also work for contacting the `auth` [Service](/reference/primitives/app/service):
-
-* `http://auth:5000` for [Services](/reference/primitives/app/service) on the same app.
-* `http://auth.myapp:5000` for other [Apps](/reference/primitives/app) on the same Rack.


docs/configuration/service-discovery.md
```diff
@@ -62,9 +62,3 @@ The `web` [Service](/reference/primitives/app/service) could reach the `auth` [S
 > Note that the internal port of the `auth` [Service](/reference/primitives/app/service) is not receiving
 > automatic SSL termination. If you want this connection to be encrypted you would need to handle SSL
 > inside the [Service](/reference/primitives/app/service).
-
-DNS search suffixes are automatically configured for internal hostnames on a Rack. The following URLs would
-also work for contacting the `auth` [Service](/reference/primitives/app/service):
-
-* `http://auth:5000` for [Services](/reference/primitives/app/service) on the same app.
```
slaskis, line 69: this one works, or are you going to remove the support of this?

Twsouza: It doesn't work, we don't have this feature. We are updating the documentation.

slaskis: strange, we use it in most of our apps. its the replacement of the links feature in v2

Twsouza: the `http://auth.myapp.convox.local:5000` option is working and not being removed.
can you share with me your rack version, if you have `internal` set to true and how are you using these DNS suffixes for your internal services?

slaskis: we're running 3.5.7-arm64 on one production rack where we do this. 

the relevant part of `convox.yml` looks like this:

```
  services:
    api:
      build: backend
      command: api
      domain: ${API_DOMAIN}
      environment:
        - CORE_URL=http://core:4000
      port: 4000
    core:
      build: backend
      command: core
      port: 4000
      internal: true
```

Twsouza: and on `api` service you can access `core` with `http://auth:5000`? or do access from another app using `http://auth.myapp:5000`?

nightfury1204 (approved):

docs/configuration/service-discovery.md
slaskis: Both

PR #618, opened 2023-04-18
Only use rack_name on a supported version

Twsouza:
### What is the feature/fix?

`rack_name` is only supported on 3.11.2+. CLI will only attach rack_name if it's on a supported version

No, it's a fix for CLI

### How to use/test it?

Install the CLI with the RC version and install a rack in a previous version: `convox rack install aws v3-8-5 --version=3.8.5`

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/rack/terraform.go
+++ b/pkg/rack/terraform.go
@@ -19,8 +19,13 @@ import (
 	"github.com/convox/stdcli"
 )
 
-const MINOR_TELEMETRY_SUPPORTED = 12
-const PATCH_TELEMETRY_SUPPORTED = 1
+const (
+	MINOR_TELEMETRY_SUPPORTED = 12
+	PATCH_TELEMETRY_SUPPORTED = 1
+
+	MINOR_RACK_NAME_SUPPORT = 11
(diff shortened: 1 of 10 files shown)


pkg/rack/terraform.go
```diff
@@ -196,7 +201,23 @@ func (t Terraform) Metadata() (*Metadata, error) {
...
+	c, err := t.Client()
+	if err == nil {
+		s, _ := c.SystemGet()
+
+		if s.Name != "" {
+			rv, _ := ConvertToReleaseVersion(s.Version)
+
+			if rv != nil {
+				if rv.Minor > MINOR_VERSION_RACK_NAME_SUPPORT ||
+					(rv.Minor == MINOR_VERSION_RACK_NAME_SUPPORT && rv.Revision >= PATCH_VERSION_RACK_NAME_SUPPORT) {
```
nightfury1204, line 215: i can see this check is occurring in several places. i think it would be better to put this check into a helper/util function

nightfury1204 (approved):

PR #621, opened 2023-04-26
Add spot and on_demand instance count in telemetry

nightfury1204:
### What is the feature/fix?

https://app.asana.com/0/1203637156732418/1204423756010922/f

** Describe the changes and if it has any breaking changes in any feature **

### How to use/test it?

** Describe how to test or use the feature **

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/heartbeat.go
+++ b/provider/aws/heartbeat.go
@@ -1,9 +1,12 @@
 package aws
 
 import (
+	"context"
 	"strings"
 
 	"github.com/convox/convox/pkg/common"
+	"github.com/pkg/errors"
+	am "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func (p *Provider) Heartbeat() (map[string]interface{}, error) {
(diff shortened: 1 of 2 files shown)


provider/k8s/k8s.go
```diff
@@ -240,13 +240,27 @@ func (p *Provider) heartbeat() error {
 		return errors.WithStack(err)
 	}
 
+	spotCnt := 0
+	onDemandCnt := 0
+
+	for i := range ns.Items {
+		switch ns.Items[i].Annotations["eks.amazonaws.com/capacityType"] {
```
yuriadams, line 247: @nightfury1204, each engine (aws, gcp, etc) as your own "Heartbeat" function that track fields specific for the provider. As sport and on-demand are related to aws provider, I think you should include this logic inside of the aws file. `/provider/aws/heartbeat.go`

yuriadams (changes requested):

Twsouza (changes requested):
I'm seeing a lot of commits and changes not related to the PR, could you confirm if is it correct?\

nightfury1204:
> I'm seeing a lot of commits and changes not related to the PR, could you confirm if is it correct?\

this depends on the yuri telemetry repo. we remove that from the staging, since that wasn't ready for the release.
i will mark this as draft for now

provider/aws/heartbeat.go
```diff
@@ -12,9 +15,28 @@ func (p *Provider) Heartbeat() (map[string]interface{}, error) {
 		return nil, err
 	}
 
+	ns, err := p.Cluster.CoreV1().Nodes().List(context.TODO(), am.ListOptions{})
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	spotCnt := 0
```
yuriadams, line 23: hey @nightfury1204, I think to avoid logic duplication, you can put this code just in > providers/k8s/k8s.go. Or create a different file and call inside `heartbeat()` function.

yuriadams (changes requested):

provider/aws/heartbeat.go
nightfury1204: provider wise it's different so it will be here. I will put this into different function.

yuriadams (approved):

PR #627, opened 2023-05-01
Fix Bug with CLI versions and rack versions. Few amends on how the event is sent to segment.

yuriadams:


pkg/rack/terraform.go
```diff
@@ -441,6 +444,13 @@ func (t Terraform) update(release string, vars map[string]string) error {
 
 	tf := filepath.Join(dir, "main.tf")
 
+	rv, _ := convertToReleaseVersion(release)
+	if rv != nil {
+		if rv.Minor > MINOR_TELEMETRY_SUPPORTED || (rv.Minor == MINOR_TELEMETRY_SUPPORTED && rv.Revision >= PATCH_TELEMETRY_SUPPORTED) {
```
Twsouza, line 449: Could you please convert this to a helper function? It's useful and we may use it in the future.

provider/k8s/telemetry.go
```diff
@@ -0,0 +1,57 @@
...
+	toSync := map[string]interface{}{}
+	for k, v := range trp.Data {
+		if !strings.Contains(skipParams, k) {
+			if strings.Contains(redactedParams, k) {
+				toSync[k] = hashParamValue(v)
+			} else {
+				toSync[k] = v
+			}
+		}
+	}
```
Twsouza, line 48: we can use an exclusion if to simplify the code
```suggestion
	for k, v := range trp.Data {
        if strings.Contains(skipParams, k) {
            continue
        }

        if strings.Contains(redactedParams, k) {
            v = hashParamValue(v)
        }

        toSync[k] = v
    }
```

Twsouza (changes requested):
just minor changes

pkg/rack/terraform.go
```diff
@@ -460,6 +467,17 @@ func (t Terraform) update(release string, vars map[string]string) error {
...
+			return true
+		}
+	}
+
+	return false
+}
+
```
Twsouza, line 480: can you move it to a file named helper?

Twsouza (approved):

PR #639, opened 2023-05-16
Add default instance type to launch template

Twsouza:
### What is the feature/fix?

Add default instance type to launch template specifications.

No

### How to use/test it?

Install a rack with the RC installed (to be created)

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated


nightfury1204 (approved):

terraform/cluster/aws/main.tf
```diff
@@ -155,6 +155,8 @@ resource "aws_launch_template" "cluster" {
     }
   }
 
+  instance_type = var.node_type
```
nightfury1204, line 158: will it cause issue if multiple node type specified
```
Node instance type. You can also pass a comma separated list of instance types
```

Twsouza: node_type is type of string, it only accepts one instance type.

image

I can update the documentation to say that is only one instance type.

Twsouza: See https://github.com/convox/convox/blob/55f4401b13aff7b23901c9294dbb5684060071a6/terraform/cluster/aws/variables.tf#L85

PR #643, opened 2023-05-22
certs renew command and fix certs list cmd

yuriadams:
### What is the feature/fix?

Task: https://app.asana.com/0/1203637156732418/1204252747981341/f

No

### How to use/test it?

$ convox certs renew -a appName


Twsouza (approved):

nightfury1204 (changes requested):
deleting secret is not the recommended way
Screenshot 2023-06-08 at 12 09 43 p m

ref: https://cert-manager.io/docs/usage/certificate/

PR #669, opened 2023-07-12
Allow for different AMI types for build node

beastawakens:
Use case where the build node can be a non-GPU instance to build the image, but the application runs on a GPU instance means we should allow for different AMI types for the build node cluster.  
Probably less of a use-case between ARM/non-ARM clusters but if someone wants to do that, why not? 😆


beastawakens:
> Thanks for the pr @beastawakens
> can you rebase it with staging?

@nightfury1204 Done!

PR #751, opened 2024-02-23
First stab at distinct health check ports

beastawakens:
### What is the feature/fix?

Allows to define a health check and liveness port separate from the main service port.  This is useful in situations where the service is listening on one port which you want to expose via a load balancer, so you define that port under the `ports` list, but this would mean the health/liveness check would not run as it relies on the port being defined under `port`.
It would also allow for a service that exposes multiple ports but you would wish to keep the health/liveness port not exposed beyond the rack, so you define that separately.

I don't believe so.  The manifest is setup to default the port to the service port if it is not defined already.

### How to use/test it?

Specify a port under the the health/liveness sections of your convox.yml

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -151,6 +151,15 @@ func (m *Manifest) ApplyDefaults() error {
 			m.Services[i].Health.Path = "/"
 		}
 
+		if s.Health.Port.Port == 0 && s.Port.Port > 0 {
+			m.Services[i].Health.Port.Port = s.Port.Port
+			if s.Port.Scheme != "" {
+				m.Services[i].Health.Port.Scheme = s.Port.Scheme
+			} else {
+				m.Services[i].Health.Port.Scheme = "http"
+			}
+		}
+
 		if s.Health.Interval == 0 {
 			m.Services[i].Health.Interval = 5
 		}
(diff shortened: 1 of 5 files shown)


nightfury1204:
@beastawakens thanks for the pr. We will look into it

pkg/manifest/manifest.go
```diff
@@ -151,6 +151,15 @@ func (m *Manifest) ApplyDefaults() error {
...
 
+		if s.Health.Port.Port == 0 && s.Port.Port > 0 {
+			m.Services[i].Health.Port.Port = s.Port.Port
+			if s.Port.Scheme != "" {
+				m.Services[i].Health.Port.Scheme = s.Port.Scheme
+			} else {
+				m.Services[i].Health.Port.Scheme = "http"
+			}
+		}
+
```
nightfury1204, line 162: What will happen if health port is defined but health scheme is not defined?

beastawakens: As the code shows, if you don't define the scheme, it will default to http.

nightfury1204: I am asking when **health port is defined** case. i can  see in the code if health port defined the it will not execute the scheme condition in here

beastawakens: Oh, I see, sorry I misunderstood 😆 Yeah, I may have missed that... let me update!

beastawakens: @nightfury1204 added default now, thank you!

PR #781, opened 2024-05-30
Add zero down node upgrade strategy

nightfury1204:
### What is the feature/fix?

This will gracefully schedule pod during node upgrade to avoid downtime


ndbroadbent:
Hi @nightfury1204, I was excited to see this in the release notes. Will this make it possible to have zero downtime during k8s version upgrades? Do I need to set a new rack param? Thanks!

beastawakens:
> Hi @nightfury1204, I was excited to see this in the release notes. Will this make it possible to have zero downtime during k8s version upgrades? Do I need to set a new rack param? Thanks!

It doesn't look like a rack parm is needed! And I agree this is very exciting, have been waiting a while for this 😁

nightfury1204:
@ndbroadbent @beastawakens yes even in k8s update it will be zero down time, but after this release
also as long as your app's following configuration doesn't allow zero replica:
```
deployment:
  minimum: 50 // minimum percentage of replica required to be up, if it's 0 then there will be downtime
  maximum: 200
```

PR #784, opened 2024-06-14
Adding cli rack install feature

numcys:
Adding Feature for installing racks to console directly through cli command by providing a runtimeid and region.

### DIFF
--- a/pkg/cli/helpers.go
+++ b/pkg/cli/helpers.go
@@ -189,7 +189,7 @@ func watch(fn func(r sdk.Interface, c *stdcli.Context) error) func(sdk.Interface
 }
 
 func checkRackNameRegex(name string) error {
-	if !regexp.MustCompile(`^[a-z0-9-]+$`).MatchString(name) {
+	if !regexp.MustCompile(`^[a-z0-9/-]+$`).MatchString(name) {
 		return fmt.Errorf("only lowercase alphanumeric characters and hyphen allowed and must not start with hyphen")
 	}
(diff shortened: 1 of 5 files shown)


pkg/cli/helpers.go
```diff
@@ -189,7 +189,7 @@ func watch(fn func(r sdk.Interface, c *stdcli.Context) error) func(sdk.Interface
 }
 
 func checkRackNameRegex(name string) error {
-	if !regexp.MustCompile(`^[a-z0-9-]+$`).MatchString(name) {
+	if !regexp.MustCompile(`^[a-z0-9/-]+$`).MatchString(name) {
```
nightfury1204, line 192: why are you allowing `/` in here?

numcys: for installing console rack we use a combination of orgname/rackname.
As the '/' wasn't allowed so it would not install console rack from cli.

nightfury1204: this will allow rack name to have `/` which we don't allow, so remove this.
Also you shouldn't mark comment resolve, I will mark them resolved if they are resolved.

numcys: I have updated the code with the changes, pls check.

nightfury1204 (changes requested):

nightfury1204 (approved):

nightfury1204 (approved):

nightfury1204 (dismissed):

PR #785, opened 2024-06-19
Added Runtimes cmd for fetching list of runtimes for a organisation

numcys:
Added cmd for fetching list of runtimes for a organisation.

PR #787, opened 2024-06-26
3.18.6

nightfury1204:
### What is the feature/fix?

- #784 
- #785 
- #786


ntner (approved):
- Install at new version
<br>

- Console CLI Rack Install Testing
- k8s API testing

PR #790, opened 2024-07-04
 Add support of using pod identity for timer

numcys:
### What is the feature/fix?
This fix ensures that timer uses pod identity of its associated service for secure resource access on promoting release.

** Describe the changes and if it has any breaking changes in any feature **
There are no breaking changes introduced with this fix.


docs/reference/cli/rack.md
```diff
@@ -27,18 +27,35 @@ Get information about the rack
...
     $ convox rack install local dev
     ...
 
     $ convox rack install aws production region=eu-west-1 node_type=t3.large
     ...
+
+    $ convox rack install aws my-org/staging region=us-east-1 --runtime=20e58437-fab7-4124-aa5a-2e5978f1149e
+    ...
 ```
+
```
nightfury1204, line 58: why these changes are in this pr?

nightfury1204 (changes requested):

docs/reference/cli/rack.md
numcys: I created the fix branch from master, staging did not had this docs change.

PR #792, opened 2024-07-09
Adding ECR image scan rack parameter

numcys:
Added Rack parameter ecr_scan_on_push_enable which can be used to change Image scan feature for images on ECR repository.

How to use:
`convox rack install <provider> <rackName> ecr_scan_on_push_enable=true`

### DIFF
--- a/provider/aws/app.go
+++ b/provider/aws/app.go
@@ -24,6 +24,9 @@ func (p *Provider) AppCreate(name string, opts structs.AppCreateOptions) (*struc
 
 	res, err := p.ECR.CreateRepository(&ecr.CreateRepositoryInput{
 		RepositoryName: aws.String(fmt.Sprintf("%s%s", p.RepositoryPrefix(), name)),
+		ImageScanningConfiguration: &ecr.ImageScanningConfiguration{
+			ScanOnPush: aws.Bool( p.ecr_scan_on_push_enable == "true"),
+		},
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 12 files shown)


provider/aws/aws.go
```diff
@@ -32,6 +32,8 @@ type Provider struct {
 	EncryptionKey string
 	Region        string
 
+	ecr_scan_on_push_enable string
```
nightfury1204, line 35: why are using snakecase here
golang we don't use snakecase
change it

provider/aws/app.go
```diff
@@ -24,6 +24,9 @@ func (p *Provider) AppCreate(name string, opts structs.AppCreateOptions) (*struc
 
 	res, err := p.ECR.CreateRepository(&ecr.CreateRepositoryInput{
 		RepositoryName: aws.String(fmt.Sprintf("%s%s", p.RepositoryPrefix(), name)),
+		ImageScanningConfiguration: &ecr.ImageScanningConfiguration{
+			ScanOnPush: aws.Bool( p.ecr_scan_on_push_enable == "true"),
```
nightfury1204, line 28: use the ecr scan variable as bool instead of string, and do the check on the var assignment

nightfury1204 (changes requested):

provider/aws/aws.go
```diff
@@ -49,11 +52,17 @@ func FromEnv() (*Provider, error) {
 		return nil, err
 	}
 
+	EcrScanOnPushEnable, err := strconv.ParseBool(os.Getenv("ECR_SCAN_ON_PUSH_ENABLE"))
```
nightfury1204, line 55: local variable always start with lowercase
update it

nightfury1204 (changes requested):

nightfury1204 (approved):

PR #796, opened 2024-07-18
Adding support to use docker creds in convox build and services pod

numcys:
### What is the feature/fix?
Adding support for using convox registries in convox build and services.

### DIFF
--- a/provider/k8s/k8s.go
+++ b/provider/k8s/k8s.go
@@ -53,6 +53,8 @@ type Provider struct {
 	ConvoxDomainTLSCertDisable       bool
 	CertManagerClient                cmclient.Interface
 	DiscoveryClient                  discovery.DiscoveryInterface
+	DockerUsername                   string
+	DockerPassword                   string
 	Domain                           string
 	DomainInternal                   string
 	DynamicClient                    dynamic.Interface
@@ -163,7 +165,7 @@ func FromEnv() (*Provider, error) {
 		MetricsClient:                    mc,
 		Name:                             ns.Labels["rack"],
 		Namespace:                        ns.Name,
-		PdbDefaultMinAvailablePercentage: common.CoalesceString(os.Getenv("PDB_DEFAULT_MIN_AVAILABLE_PERCENTAGE"),"50"),
+		PdbDefaultMinAvailablePercentage: common.CoalesceString(os.Getenv("PDB_DEFAULT_MIN_AVAILABLE_PERCENTAGE"), "50"),
 		Password:                         os.Getenv("PASSWORD"),
 		Provider:                         common.CoalesceString(os.Getenv("PROVIDER"), "k8s"),
 		RackName:                         rn,
@@ -175,6 +177,8 @@ func FromEnv() (*Provider, error) {
 		SubnetIDs:                        os.Getenv("SUBNET_IDS"),
 		Version:                          common.CoalesceString(os.Getenv("VERSION"), "dev"),
 		VpcID:                            os.Getenv("VPC_ID"),
+		DockerUsername:                   os.Getenv("DOCKER_HUB_USERNAME"),
+		DockerPassword:                   os.Getenv("DOCKER_HUB_PASSWORD"),
 	}
 
 	p.RdsProvisioner = rds.NewProvisioner(p)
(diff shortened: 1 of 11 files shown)


provider/k8s/app.go
```diff
@@ -52,6 +52,19 @@ func (p *Provider) AppCreate(name string, opts structs.AppCreateOptions) (*struc
 		return nil, errors.WithStack(err)
 	}
 
+	sourceSecret, _ := p.Cluster.CoreV1().Secrets(p.RackName+"-system").Get(context.TODO(), "registries", am.GetOptions{})
```
nightfury1204, line 55: why are you ignoring error here? you should not ignore error

also use `p.Namespace` for system namespace

provider/k8s/app.go
```diff
@@ -52,6 +52,19 @@ func (p *Provider) AppCreate(name string, opts structs.AppCreateOptions) (*struc
...
+	if sourceSecret != nil {
+		newSecret := &ac.Secret{
+			ObjectMeta: am.ObjectMeta{
+				Name:      sourceSecret.Name,
+				Namespace: p.AppNamespace(name),
+			},
+			Type: sourceSecret.Type,
+			Data: sourceSecret.Data,
+		}
+		p.Cluster.CoreV1().Secrets(p.AppNamespace(name)).Create(context.TODO(), newSecret, am.CreateOptions{})
```
nightfury1204, line 65: why are you ignoring error?

terraform/api/k8s/main.tf
```diff
@@ -106,6 +106,10 @@ resource "kubernetes_deployment" "api" {
           }
         }
 
+        image_pull_secrets {
+          name = "registries"
```
nightfury1204, line 110: why did explicitly specify `registries` here?

nightfury1204 (changes requested):
Hey why you did not choose docker hub authentication here?

provider/k8s/process.go
```diff
@@ -573,6 +575,34 @@ func (p *Provider) podSpecFromRunOptions(app, service string, opts structs.Proce
 		s.Containers[0].Resources.Limits["memory"] = resource.MustParse(fmt.Sprintf("%dMi", *opts.MemoryLimit))
 	}
 
+	p.DockerUsername = common.CoalesceString(os.Getenv("DOCKER_HUB_USERNAME"),"")
+	p.DockerPassword = common.CoalesceString(os.Getenv("DOCKER_HUB_PASSWORD"),"")
```
nightfury1204, line 579: do not initialize the var here. you should initialize that in here: https://github.com/convox/convox/blob/master/provider/k8s/k8s.go#L94

provider/k8s/process.go
```diff
@@ -573,6 +575,34 @@ func (p *Provider) podSpecFromRunOptions(app, service string, opts structs.Proce
...
+
+	_, err = p.CreateOrPatchSecret(p.ctx, am.ObjectMeta{
```
nightfury1204, line 581: add a check if docker username and password is not provided, don't execute these logic

nightfury1204 (changes requested):
I don't see image pull secrets related changes for atom and resolver here, is it already provided?

numcys:
Yes Nahid, Atom and resolver already are provided.

provider/k8s/k8s.go
```diff
@@ -151,6 +153,8 @@ func FromEnv() (*Provider, error) {
 		ConvoxDomainTLSCertDisable: os.Getenv("CONVOX_DOMAIN_TLS_CERT_DISABLE") == "true",
 		CertManagerClient:          cm,
 		DiscoveryClient:            kc.Discovery(),
+		DockerUsername: 			common.CoalesceString(os.Getenv("DOCKER_HUB_USERNAME"),""),
+		DockerPassword: 			common.CoalesceString(os.Getenv("DOCKER_HUB_PASSWORD"),""),
```
nightfury1204, line 157: common.CoalesceString is not needed here
currently it is like if the value is empty then set empty.
fix this please

nightfury1204 (changes requested):

provider/k8s/k8s.go
numcys: Added the changes...

nightfury1204 (dismissed):

PR #797, opened 2024-07-22
Add support to configure PDB default min availability using rack param

numcys:
### What is the feature/fix?
Adding Rack param `pdb_default_min_available_percentage`  for setting the Default minimum percentage of Pod disruptions Budget for the Rack.


provider/k8s/controller_deployment.go
```diff
@@ -154,9 +155,9 @@ func (c *DeployController) SyncPDB(d *apps.Deployment, remove bool) error {
 		remove = true
 	}
 
-	minAvailable := &intstr.IntOrString{
+	pdb_default_min_available_percentage := &intstr.IntOrString{
 		Type:   intstr.String,
-		StrVal: "50%",
+		StrVal: os.Getenv("PDB_DEFAULT_MIN_AVAILABLE_PERCENTAGE")+"%",
```
nightfury1204, line 160: use this pattern for this: https://github.com/convox/convox/blob/master/provider/k8s/k8s.go#L157

don't take the value from env directly here.
also it's better to put validation on the value

terraform/api/aws/main.tf
```diff
@@ -39,18 +39,19 @@ module "k8s" {
...
+    EFS_FILE_SYSTEM_ID                   = var.efs_file_system_id
+    BUILD_DISABLE_CONVOX_RESOLVER        = var.build_disable_convox_resolver
+    PROVIDER                             = "aws"
+    RESOLVER                             = var.resolver
+    ROUTER                               = var.router
+    SOCKET                               = "/var/run/docker.sock"
+    ECR_SCAN_ON_PUSH_ENABLE              = var.ecr_scan_on_push_enable
+    SUBNET_IDS                           = join(",", var.subnets)
+    VPC_ID                               = var.vpc_id
+    PDB_DEFAULT_MIN_AVAILABLE_PERCENTAGE = var.pdb_default_min_available_percentage
```
nightfury1204, line 55: make it lexicographically sorted

terraform/api/aws/variables.tf
```diff
@@ -100,3 +100,8 @@ variable "subnets" {
 variable "vpc_id" {
   type = string
 }
+
+variable "pdb_default_min_available_percentage" {
```
nightfury1204, line 104: same here make it lexicographically sorted

provider/k8s/controller_deployment.go
numcys: Added field in Provider struct for the variable and applied validation for a default value in case the env variable is nil.

terraform/api/aws/main.tf
numcys: Sorted Lexicographically.

terraform/api/aws/variables.tf
numcys: Sorted the variable and other ones too lexicographically.

provider/k8s/controller_deployment.go
```diff
@@ -154,9 +156,11 @@ func (c *DeployController) SyncPDB(d *apps.Deployment, remove bool) error {
 		remove = true
 	}
 
-	minAvailable := &intstr.IntOrString{
+	c.Provider.PdbDefaultMinAvailablePercentage = common.CoalesceString(os.Getenv("PDB_DEFAULT_MIN_AVAILABLE_PERCENTAGE"),"50")
```
nightfury1204, line 159: do not initialize the var here. you should initialize that in here: https://github.com/convox/convox/blob/master/provider/k8s/k8s.go#L94

provider/k8s/controller_deployment.go
```diff
@@ -154,9 +156,11 @@ func (c *DeployController) SyncPDB(d *apps.Deployment, remove bool) error {
...
+
+	pdb_default_min_available_percentage := &intstr.IntOrString{
```
nightfury1204, line 161: in golang we don't use snakecase, fix it please.

provider/k8s/controller_deployment.go
```diff
@@ -154,9 +156,11 @@ func (c *DeployController) SyncPDB(d *apps.Deployment, remove bool) error {
...
 		Type:   intstr.String,
-		StrVal: "50%",
+		StrVal: c.Provider.PdbDefaultMinAvailablePercentage+"%",
```
nightfury1204, line 163: it is preferable to `fmt.Sprintf` here

nightfury1204 (changes requested):

provider/k8s/controller_deployment.go
numcys: Corrected this, will this in my mind.

provider/k8s/controller_deployment.go
numcys: Used fmt.sprintf here...

nightfury1204 (changes requested):

nightfury1204 (dismissed):

PR #802, opened 2024-08-06
Added Rack params for increasing kubelet image pull limits  - Feature/kubelet pull limits params

numcys:
### What is the feature/fix?
Added two rack params - registryPullQPS and registryBurst to configure image pull limits in kubelet configuration.


terraform/cluster/aws/main.tf
```diff
@@ -278,6 +278,42 @@ resource "aws_launch_template" "cluster" {
     }
   }
 
+  user_data = base64encode(<<-EOF
```
nightfury1204, line 281: use terraform template file instead

also this will trigger update for all the existing cluster

can you change it as we discussed like if user explicitly don't set it then no need to provide it as userdata?

terraform/system/aws/variables.tf
```diff
@@ -229,6 +229,16 @@ variable "region" {
   default = "us-east-1"
 }
 
+variable "registry_pull_qps" {
```
nightfury1204, line 232: change param names to `kubelet_registry_pull_qps` and `kubelet_registry_burst` everywhere

nightfury1204 (changes requested):

PR #807, opened 2024-08-15
Added feature for using custom image in convox resources - feature/resource-custom-image-params

numcys:
### What is the feature/fix?
We can pass image parameter in the convox.yml manifest file in resources. It allows for using custom images for the available resource types. 

How to use it:
Earlier:
```
  db:  
    type: postgres
    options:
      version: 13
```

New way:
```
  db2:
    type: postgres
    image: pgvector/pgvector:pg16
```

Note:
1. The Image tag has to be passed with the image name in case the custom image does not have latest tag available on the registry.
2. The custom image used is supposed to be compatible with the resource type. In case if it is not then the resource creation will fail.

PR #811, opened 2024-08-27
3.18.10

nightfury1204:
### What is the feature/fix?

- #810 
- #809 
- #807


ntner (approved):
- Install at new version
<br>

- Extended RDS testing
- Wildcard Certificate Testing
- Custom resource image testing

PR #821, opened 2024-09-20
Added resource quota for priority class in convox system namespace to fix gcp rack installation

numcys:
### What is the feature/fix?
This PR contains fix for failing rack creation and github action job failures on GCP platform. This is caused due to resource quota handled differently in GCP and AWS, GCP requires us to create seperate namespaced resource quota for the default priority class that k8s provides.
Added Resource quota in the convox system namespace which can be used by GKE to assign appropriate priority to pods assigned priority class during scheduling.

### DIFF
--- a/terraform/api/k8s/main.tf
+++ b/terraform/api/k8s/main.tf
@@ -3,6 +3,25 @@ resource "random_string" "password" {
   special = false
 }
 
+resource "kubernetes_resource_quota" "gcp-critical-pods" {
+  metadata {
+    name = "gcp-critical-pods"
+    namespace = var.namespace
+  }
+  spec {
+    hard = {
+      pods = "1G"
+    }
+    scope_selector {
+      match_expression {
+        scope_name = "PriorityClass"
+        operator = "In"
+        values = ["system-node-critical", "system-cluster-critical"]
+      }
+    }
+  }
+}
+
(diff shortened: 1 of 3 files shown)


terraform/system/gcp/versions.tf
```diff
@@ -3,6 +3,7 @@ terraform {
     google = {
       source  = "hashicorp/google"
       version = "~> 4.69.1"
+      configuration_aliases = [ google.direct, ]
```
nightfury1204, line 6: remove the comma here?
```
google.direct,
```

terraform/api/k8s/main.tf
```diff
@@ -3,6 +3,25 @@ resource "random_string" "password" {
...
 
+resource "kubernetes_resource_quota" "gcp-critical-pods" {
+  metadata {
+    name = "gcp-critical-pods"
+    namespace = var.namespace
+  }
+  spec {
+    hard = {
+      pods = "1G"
+    }
```
nightfury1204, line 14: Why it is `1G`?

numcys: 1G is the default hard limit on the Quota in the other namespace, I kept it identical in the convox system namespace also. I checked, it says 1G is 1000 pods limit.

nightfury1204 (approved):

PR #824, opened 2024-10-01
3.19.1

nightfury1204:
### What is the feature/fix?

- #823
- #822


ntner (approved):
- Install at new version
<br>

- Additional RDS testing
- CloudWatch testing
- DNS01 testing

PR #831, opened 2024-10-21
Improve buildkit multi service cache

nightfury1204:
### What is the feature/fix?

This will improve buildkit multi service cache


advance512:
Hi there @nightfury1204 , are there any plans on backporting this to v2 racks?
The caching is really problematic there.. we have a complete app rebuild every few days, which is 5-10x times longer.

nightfury1204:
> Hi there @nightfury1204 , are there any plans on backporting this to v2 racks? The caching is really problematic there.. we have a complete app rebuild every few days, which is 5-10x times longer.

@ntner can you look into this?

PR #832, opened 2024-11-04
Introduce custom user data for launch templates

beastawakens:
### What is the feature/fix?

We have a use-case where we need to run user-data on our v3 Rack (specifically, mounting and attaching the instance-attached storage automatically).

This PR adds a `user_data` rack parameter which will percolate through and changes the name of the `kublet_config_override.sh` script to a more generic name and adds user data to that if is specified.

No

### How to use/test it?

Add some `user_data` rack parameter and check that it is run on instance startup.

### Checklist
- [N/A] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [x] Documentation updated


beastawakens:
@nightfury1204 this would be super lovely and useful to get into the main release 😉

beastawakens:
> > @nightfury1204 this would be super lovely and useful to get into the main release 😉
> 
> Thanks for pr @beastawakens. I asked @numcys to look into it and enhance it if possible like giving large scripts as string will be problematic, so I am wondering if it's possible to give http accessible script file url as param and it will fetch and add the scripts from the url.

That would be great as an option for sure, the v2 Racks support single line scripts (with `InstanceBootCommand` and `InstanceRunCommand`) and we can always combine simple multi-line commands into one.

PR #833, opened 2024-11-06
adding disable_kernel_mitigations rack param

numcys:
### What is the feature/fix?
Added Rack parameter `disable_kernel_mitigations` to disable the kernel mitigations on all nodes. 
Note-
1. Enabling this disabled mitigations which can improve performance but also introduce new security vulnerabilities.
2. Enabling this param recreates all nodes.

Usage:
```
convox rack params set disable_kernel_mitigations=true
```


beastawakens:
@numcys see also #832 for work in the same area!

PR #840, opened 2024-11-22
Feature/multiline annotations

numcys:
### What is the feature/fix?
Added Support for adding multiline annotations to deployments. 
example manifest:
```
services:
  web:
    build: .
    command: node app.js
    port: 3000
    annotations:
        - "key1=value1"
        - key2: "value2"
        - key3: |
            this is it
        - ad.datadoghq.com/test.checks: |
            {   
                "my-test": {
                "init_config": {
                    "is_jmx": true,
                    "collect_default_metrics": true
                },
                "instances": [{
                    "host": "localhost",
                    "port": "1234"
                }]
            }
            }
```

### DIFF
--- a/pkg/common/format.go
+++ b/pkg/common/format.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"bytes"
+	"html"
 	"regexp"
 
 	yaml "gopkg.in/yaml.v2"
@@ -12,6 +13,10 @@ var (
 )
 
 func FormatYAML(data []byte) ([]byte, error) {
+
+	// Unescape HTML entities first
+	data = []byte(html.UnescapeString(string(data)))
+
 	ps := yamlSplitter.Split(string(data), -1)
 	bs := make([][]byte, len(ps))
 
--- a/pkg/manifest/service.go
+++ b/pkg/manifest/service.go
@@ -2,6 +2,7 @@ package manifest
 
 import (
 	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strings"
@@ -57,8 +58,8 @@ type Service struct {
 type Annotations []Annotation
 
 type Annotation struct {
-	Key   string
-	Value string
+	Key   string      `yaml:"key"`
+	Value interface{} `yaml:"value"`
 }
 
 type InitContainer struct {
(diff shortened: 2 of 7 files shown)


pkg/common/format.go
```diff
@@ -12,6 +13,10 @@ var (
 )
 
 func FormatYAML(data []byte) ([]byte, error) {
+
+	// Unescape HTML entities first
+	data = []byte(html.UnescapeString(string(data)))
```
nightfury1204, line 18: Why this is needed here?

pkg/manifest/service.go
```diff
@@ -54,6 +55,13 @@ type Service struct {
 	AccessControl      AccessControlOptions  `yaml:"accessControl,omitempty"`
 }
 
+type Annotations []Annotation
+
+type Annotation struct {
+	Key   string      `yaml:"key"`
+	Value interface{} `yaml:"value"`
```
nightfury1204, line 62: why this is interface instead of string?

pkg/manifest/service.go
```diff
@@ -335,18 +343,36 @@ func (sr ServiceResource) GetConfigMapKey() string {
...
-		annotations[parts[0]] = parts[1]
+		jsonStr := string(jsonValue)
+
+		if containsJSONStructure(jsonStr) {
+			key := fmt.Sprintf("%s : |", annotation.Key)
+			annotations[key] = jsonStr
+		} else {
+			key := fmt.Sprintf("%s :", annotation.Key)
+			annotations[key] = jsonStr
+		}
```
nightfury1204, line 366: why json marshal and json str here?

pkg/manifest/yaml.go
```diff
@@ -158,6 +159,75 @@ func (v *Service) SetName(name string) error {
...
+						// Successfully unmarshaled JSON string into a map
+						annotations = append(annotations, Annotation{Key: keyStr, Value: jsonValue})
+					} else {
+						// If not a valid JSON, store the string as-is
+						annotations = append(annotations, Annotation{Key: keyStr, Value: jsonStr})
+					}
+				} else {
+					// Store as is if not a JSON string
+					annotations = append(annotations, Annotation{Key: keyStr, Value: v})
+				}
```
nightfury1204, line 200: why json unmarshal here?

provider/k8s/template.go
```diff
@@ -126,9 +129,45 @@ func (p *Provider) templateHelpers() template.FuncMap {
 		"volumeTo": func(v string) (string, error) {
 			return volumeTo(v)
 		},
+		"splitIntoLines": splitIntoLines,
 	}
 }
 
+func splitIntoLines(value interface{}) []string {
```
nightfury1204, line 136: why this is necessary?

provider/k8s/template/app/service.yml.tmpl
```diff
@@ -19,9 +19,13 @@ apiVersion: v1
...
-    {{ range keyValue .Annotations }}
-    {{.Key}}: "{{ quoteEscape .Value}}"
-    {{ end }}
+    {{- range $key, $value := .Annotations }}
+        {{ $key }}
+        {{- $lines := splitIntoLines $value }}
+        {{- range $lines }}
+          {{ . }}
+        {{- end }}
+    {{- end }}
```
nightfury1204, line 28: why this change is needed?

pkg/common/format.go
numcys: The double quotes " were being converted to html entity code &#34. so to parse that as it is in annotations I added it.

pkg/manifest/service.go
numcys: The value could be of a string as well as map, so needed to keep this as interface to handle nested json structure.

pkg/manifest/service.go
numcys: This is to format the json structure of the annotation properly.

pkg/manifest/yaml.go
numcys: If the nested json needs to be formatted, otherwise we add the value as it is.

provider/k8s/template.go
numcys: To solve issue of templating the value of interface type, I created function to split the interface into []string, and then traverse over it and template that.

provider/k8s/template/app/service.yml.tmpl
numcys: This is for the above said change of printing slice of strings for a annotation instead of interface.

nightfury1204 (changes requested):
I think it shouldn't be json specific implementaion. It should work for all types formated string.

PR #841, opened 2024-11-26
Reintroduce multiple node types

beastawakens:
### What is the feature/fix?

This reintroduces being able to run multiple instance types on an AWS rack, which was the state for a long while until launch templates were introduced.

We've run into issues fairly frequently when running spot instances where all the capacity for a specific instance is used up in a region so our Rack will disappear or be completely decimated.  If we allow for multiple instance types, it would give us more opportunity to create nodes of other types to not suffer from this!

Launch templates in Terraform only allow for one instance type to be set, so this PR creates one node group per node_type set in the rack parameter and an associated launch template for that node group.

The first node_type in the rack parameter is considered the primary type to determine the appropriate release of the API etc to use, so it would be highly recommended to keep all the instance types in the same family (at least keep them all ARM or non-ARM!).

The functionality to determine if the instance type contains a GPU or runs on an ARM processor has been moved to a specific module so that it can be more easily referenced in multiple places.

Node groups for any instance beyond the first one will have their minimum and desired instance count set to 0 so they don't all spin up instances immediately, and will allow the kube-scheduler/ASG to scale them up as needed by the apps on the rack.

The implementation of the `MIXED` capacity type has changed from setting on-demand in one AZ only, to setting on-demand for the first node type and spot for all subsequent ones.  This should be more reliable in the case of a spot shortage as it will maintain instances across 3 AZ's for redundancy rather than just in one.

There is undoubtedly other factors or things I'm missing in this PR but I hope it's a great starting off point for a more reliable underlying infrastructure.

I also implemented the same for the build node groups but if this is intended to purely by a singleton instance, maybe this is unnecessary and we should enforce the build_node_type being a single instance type. 

Change in implementation to the `MIXED` capacity type which isn't breaking but could be highlighted.

### How to use/test it?

Set `node_type` rack param to a comma-delimited list of instance types and see the node groups being created.  Check that node groups beyond the first one have a desired and minimum size of 0.  Cause scale up events on the rack and see if other instance types get scaled up as desired.

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/terraform/cluster/aws/amitype.tf
+++ b/terraform/cluster/aws/amitype.tf
@@ -1,0 +1,11 @@
+variable "node_type" {
+  type = string
+}
+
+output "gpu_type" {
+  value = substr(var.node_type, 0, 1) == "g" || substr(var.node_type, 0, 1) == "p" || substr(var.node_type, 0, 3) == "inf" || substr(var.node_type, 0, 3) == "trn"
+}
+
+output "arm_type" {
+  value = substr(var.node_type, 0, 2) == "a1" || substr(var.node_type, 0, 5) == "hpc7g" || substr(var.node_type, 0, 4) == "im4g" || substr(var.node_type, 0, 4) == "is4g" || substr(var.node_type, 2, 1) == "g"
+}
--- a/terraform/cluster/aws/main.tf
+++ b/terraform/cluster/aws/main.tf
@@ -40,6 +40,10 @@ locals {
   availability_zones     = var.availability_zones != "" ? compact(split(",", var.availability_zones)) : data.aws_availability_zones.available.names
   network_resource_count = var.high_availability ? 3 : 2
   oidc_sub               = "${replace(aws_iam_openid_connect_provider.cluster.url, "https://", "")}:sub"
+  node_type_list = split(",", var.node_type)
+  node_type_map  = zipmap(range(length(local.node_type_list)), local.node_type_list)
+  build_node_type_list = split(",", var.build_node_type)
+  build_node_type_map = zipmap(range(length(local.build_node_type_list)), local.build_node_type_list)
 }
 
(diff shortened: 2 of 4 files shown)


nightfury1204:
Thanks @beastawakens for the initiative. We have an internal ticket for that.

beastawakens:
> Thanks @beastawakens for the initiative. We have an internal ticket for that.

Awesome @nightfury1204 , this would be a huge thing for us (and probably for many customers)

- We currently see big disruption occasionally on our mixed/spot racks when someone buys up all the spot capacity for an instance type in a region/AZ.  If we could effectively specify multiple instance types again, that would immediately solve that.
- For some of our apps which rely on mixed needs for GPU/non-GPU purposes, for cost-efficiency we have to split them over two racks and then communicate internally with VPC peering and an internal custom load balancer which is a lot of overhead that could be solved easily by allowing mixed types in the same rack.

beastawakens:
@nightfury1204 Just wondering if there's any rough expectation of when having the multi-node types again might land?  If there's anything I can do to assist then please let me know, this is a key thing for a lot of people 😊

nightfury1204:
> @nightfury1204 Just wondering if there's any rough expectation of when having the multi-node types again might land? If there's anything I can do to assist then please let me know, this is a key thing for a lot of people 😊

Hey sorry for the late reply. I was on vacation. 

@ntner can you look into this?

beastawakens:
Any progress @ntner ? Or anything I can do to help?

ntner:
Hey @beastawakens we're pushing a large update next week, and then this will be included in the following update.  We are reworking this some to add service node selector and additional functionality before release.

nightfury1204:
Closing it on behalf of this: #867

PR #842, opened 2024-11-26
Add support for multiline annotations

nightfury1204:
### What is the feature/fix?

This will allow to define multi line service annotation

```
services:
  web:
    build: .
    port: 3000
    annotations:
        - "key1=value1"
        - k2 : "v2"
        - "simple.key=value"
        - "quoted.key=\"quoted value\""
        - another.key: "another value"
        - multi.line.test: |
            This is line 1
            Line 2 with a "quote" and a special character: $
```

### DIFF
--- a/pkg/manifest/service.go
+++ b/pkg/manifest/service.go
@@ -17,7 +17,7 @@ type Service struct {
 	Name string `yaml:"-"`
 
 	Agent              ServiceAgent          `yaml:"agent,omitempty"`
-	Annotations        ServiceAnnotations    `yaml:"annotations,omitempty"`
+	Annotations        Annotations           `yaml:"annotations,omitempty"`
 	Build              ServiceBuild          `yaml:"build,omitempty"`
 	Certificate        Certificate           `yaml:"certificate,omitempty"`
 	Command            string                `yaml:"command,omitempty"`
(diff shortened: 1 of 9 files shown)


pkg/manifest/service.go
```diff
@@ -335,13 +342,11 @@ func (sr ServiceResource) GetConfigMapKey() string {
 	return DEFAULT_RESOURCE_ENV_NAME
 }
 
-// skipcq
 func (s Service) AnnotationsMap() map[string]string {
 	annotations := map[string]string{}
 
 	for _, a := range s.Annotations {
-		parts := strings.SplitN(a, "=", 2)
```
ndbroadbent, line 343: Ahh, I just upgraded Convox to the latest stable version and was wondering why my proxy-buffer-size annotations dissappeared!

```
    ingressAnnotations:
      # Required for link headers on homepage for preloading assets
      - nginx.ingress.kubernetes.io/proxy-buffer-size=32k
```

so this needs to be key/value now, like this?

```
    ingressAnnotations:
      # Required for link headers on homepage for preloading assets
      - nginx.ingress.kubernetes.io/proxy-buffer-size: 32k
```

ndbroadbent: Ahh, I got it work again by renaming `ingressAnnotations` to `annotations`. Sorry I missed that:

```
services:
  web:
    annotations:
      - nginx.ingress.kubernetes.io/proxy-buffer-size: "32k"
```

PR #865, opened 2025-02-20
Upgrade to k8s 1.30 

nightfury1204:
### What is the feature/fix?

 - update k8s version to 1.30
 - update docker buildkit version to v0.19.0
 - update cluster autoscaler to v1.32.0
- update metric server to 0.7.2
 - update nginx controller to 1.12.0 (server snippet and configuration snippet now begin flagged if risk is Critical, so if you use custom snippet make sure it is not flagged as critical risk)
-  update ebs driver to eks managed v1.39.0
-  update efs driver to v2.1.4
-  update coredns to v1.11.4
-  update kube proxy to v1.30.9
-  update vpc cni to v1.19.2


ntner (approved):
- Install at new version
<br>

- ebs/efs volume testing
- ingress and service communication testing
- ssh and instance testing
- Quick Apps testing

PR #860, opened 2025-02-04
Add support for custom AMI on EKS racks

beastawakens:
### What is the feature/fix?

Adds support for custom AMI support on v3/EKS racks.  Also allows for distinct custom AMI support for the build node if needed.

**To use custom ami in build node you have to explicitly define the build node ami id. It won't use node ami id.**

### Does it has a breaking change?

No

### How to use/test it?

Specify the `ami_id` rack parameter with a valid AMI ID to use that as your base node image.
Optionally also specify a `build_ami_id` rack parameter with a valid AMI ID to further use that as the node image for your build node.

### Checklist
- [N/A] New coverage tests
- [N/A] Unit tests passing
- [x] E2E tests passing
- [ ] E2E downgrade/update test passing
- [x] Documentation updated
- [x] No warnings or errors on Deepsource/Codecov

PR #869, opened 2025-03-12
Revert "Add support for custom AMI on EKS racks"

nightfury1204:
Reverts convox/convox#860


nightfury1204:
@beastawakens reverting your pr, since it is not complete and properly tested. Custom ami doesn't allow k8s version to be specified. Custom ami doesn't auto join eks cluster.

beastawakens:
> @beastawakens reverting your pr, since it is not complete and properly tested. Custom ami doesn't allow k8s version to be specified. Custom ami doesn't auto join eks cluster.

@nightfury1204 oh damn 😞 that's odd, I'll try and take a look into it if you don't get a chance to.  This would be a great feature still.

nightfury1204:
> @nightfury1204 oh damn 😞 that's odd, I'll try and take a look into it if you don't get a chance to. This would be a great feature still.

@beastawakens I am working on the extended feature on top of to support additional node groups where i am trying to put support for custom ami. If that succeed, then that will solve you requirements i think.

PR #875, opened 2025-04-01
3.21.0

nightfury1204:
### What is the feature/fix?

- upgrade k8s to 1.31
- buildkit version to v0.20.1
- ebs csi driver to v1.41.0-eksbuild.1
- efs csi driver to v2.1.6-eksbuild.1
- kube proxy to v1.31.2-eksbuild.3


ntner (approved):
- Install at new version
<br>

PR #877, opened 2025-04-07
Initial fix to add readinessProbe for grpc services

beastawakens:
### What is the feature/fix?

Experienced an issue when deploying a grpc service, as it started receiving traffic before it was ready to.  It appears the grpc health check is implemented as a livenessProbe (with hardcoded default values) rather than a readinessProbe that reflects the settings in the health config from convox.yml.

This PR adds a readinessProbe and uses the values from convox.yml

There's probably a wider/bigger improvement to be made to allow any of the startup/liveness/health configs to be enabled using grpc rather than http.  I'm happy to raise a PR for that as well if desired, but wanted to get this one in quickly to unblock the immediate issue.

No

### How to use/test it?

Deploy a grpc service with the `grpcHealthEnabled: true` and a `health` configuration set in the convox.yml.  See that the readinessProbe is created to reflect the configuration set.

### Checklist
- [N/A] New coverage tests
- [N/A] Unit tests passing
- [ ] E2E tests passing
- [N/A] Documentation updated

### DIFF
--- a/provider/k8s/template/app/service.yml.tmpl
+++ b/provider/k8s/template/app/service.yml.tmpl
@@ -255,12 +255,20 @@ spec:
           successThreshold: 1
           failureThreshold: 3
         {{ else if $.Service.GrpcHealthEnabled }}
+        readinessProbe:
+          grpc:
+            port: {{.}}
+          initialDelaySeconds: {{$.Service.Health.Grace}}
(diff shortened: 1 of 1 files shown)


ntner:
@beastawakens thanks Ed we will look at this asap

nightfury1204 (approved):

nightfury1204 (approved):

PR #881, opened 2025-04-11
3.21.1

nightfury1204:
### What is the feature/fix?

- #877
- #880 
- #879 
- #878


ntner (approved):
- Install at new version
<br>

- gRPC testing
- Additional node group testing
- nVidia GPU and scale testing
- ENV and release testing

PR #888, opened 2025-05-06
3.21.3

nightfury1204:
### What is the feature/fix?

- #887 
- #886


ntner (approved):
- Install at new version
<br>

- Additional node group testing
- nVidia GPU and scale testing with `convox run`
- Liveness testing

PR #914, opened 2025-08-13
3.22.0

nightfury1204:
### What is the feature/fix?

- #913 
- #912 
- #911 
- Upgrade ami from AL2 to AL2023 #911


ntner (approved):
- Install at new version
<br>

- Build and workflows testing

PR #915, opened 2025-08-14
Allow for more flexibile instance selection in node groups

beastawakens:
### What is the feature/fix?

This allows for more flexible instance selection in additional node groups and additional build groups by utilising the "instance requirements" feature of launch templates rather than specific instance type.
In the node groups config json, `type` is now optional, but you should specify either that or both the `cpu` and `mem` fields.  That will determine which type of launch template is created.
If `cpu` and `mem` fields are set to define the minimum cpu and memory requirements of the instance, you are also able to specify the `types` field to set the "allowed instance types" configuration within the instance requirements block to further limit the instance types desired if required.

No.  Existing configurations will work just fine.

### How to use/test it?

See examples specified in the documentation changes.

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [x] Documentation updated

### DIFF
--- a/docs/configuration/rack-parameters/aws/additional_node_groups_config.md
+++ b/docs/configuration/rack-parameters/aws/additional_node_groups_config.md
@@ -26,7 +26,10 @@ The `additional_node_groups_config` parameter takes a JSON array of node group c
 
 | Field | Required | Description | Default |
 |-------|----------|-------------|---------|
-| `type` | Yes | The EC2 instance type to use for the node group |  |
+| `type` | Yes (or set the `cpu` and `mem` fields) | The EC2 instance type to use for the node group |  |
+| `cpu` | Yes (or set the `type` field) | The minimum number of vCPUs required from the node |  |
+| `mem` | Yes (or set the `type` field) | The minimum mib of memory required from the node |  |
+| `types` | No (but can use in conjunection with the `cpu` and `mem` fields) | List of instance types to apply your specified `cpu` and `mem` attributes against. All other instance types are ignored, even if they match your specified attributes. You can use strings with one or more wild cards, represented by an asterisk (*), to allow an instance type, size, or generation. |  |
 | `disk` | No | The disk size in GB for the nodes | Same as main node disk |
 | `capacity_type` | No | Whether to use on-demand or spot instances | `ON_DEMAND` |
 | `min_size` | No | Minimum number of nodes | 1 |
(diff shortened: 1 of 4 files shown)


docs/configuration/rack-parameters/aws/additional_node_groups_config.md
```diff
@@ -26,7 +26,10 @@ The `additional_node_groups_config` parameter takes a JSON array of node group c
 
 | Field | Required | Description | Default |
 |-------|----------|-------------|---------|
-| `type` | Yes | The EC2 instance type to use for the node group |  |
+| `type` | Yes (or set the `cpu` and `mem` fields) | The EC2 instance type to use for the node group |  |
+| `cpu` | Yes (or set the `type` field) | The minimum number of vCPUs required from the node |  |
+| `mem` | Yes (or set the `type` field) | The minimum mib of memory required from the node |  |
```
nightfury1204: Why are `cpu` and `memory` necessary here?
Wouldn't just `types` be enough?

nightfury1204 (changes requested):
this implementation didn't consider ami_type in the aws_eks_node_group

terraform/cluster/aws/node_groups.tf
```diff
@@ -70,7 +79,7 @@ resource "random_id" "additional_node_groups" {
 module "amitype" {
   source    = "../../helpers/aws"
   for_each  = { for ng in local.additional_node_groups_with_defaults : ng.id => ng }
-  node_type = each.value.type
+  node_type = coalesce(each.value.type, try(each.value.types[0], null))
```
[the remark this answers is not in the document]
beastawakens: It could be worth clarifying in the docs, but I would hope and assume that people wouldn't specify different incompatible instance types (x64, ARM, GPU) in the same node group 🤣
That's why I took the easy option and if `type` isn't set, to use the first entry in `types` to get the AMI type required.

docs/configuration/rack-parameters/aws/additional_node_groups_config.md
nightfury1204: nano and micro won't work considering running all those daemonsets and other. we can set that as minimum, omit cpu and mem as option. instance requirements should only be considering the allowed types in that case based on that minimum 2 cpu and 2 gib mem.

PR #932, opened 2025-10-16
Fix: Store release-cleanup timestamp in configmap

ndbroadbent:
### What is the feature/fix?

- The release-cleaner task used to record its last run in the system namespace annotation. Our Terraform module doesn’t manage that annotation, so every terraform apply overwrote it with an empty map, effectively erasing the timestamp and breaking the cleanup schedule.
- Fix: persist the timestamp in a dedicated ConfigMap (convox-release-cleanup), seeding it from the legacy annotation on first run and updating it after each cleanup. Terraform no longer wipes out the value.
- Also fixed some DeepSource warnings

It also fixes this terraform drift issue:

```
  Terraform will perform the following actions:

    # module.system.module.rack.module.k8s.kubernetes_namespace.system will be updated in-place
    ~ resource "kubernetes_namespace" "system" {
          id                               = "staging-3-17-system"
          # (1 unchanged attribute hidden)

        ~ metadata {
            ~ annotations      = {
                - "convox.io/last-release-build-cleanup" = "2025-10-16T00:21:56Z" -> null
                  # (1 unchanged element hidden)
              }
              name             = "staging-system"
              # (5 unchanged attributes hidden)
          }

          # (1 unchanged block hidden)
      }

  Plan: 0 to add, 1 to change, 0 to destroy.
```

- N/A (server-side behavior only).

### Does it have a breaking change?

- No breaking changes. Cleanup cadence and behavior remain the same, but the state lives in a ConfigMap instead of an annotation.

### How to use/test it?

1. Apply/build the updated provider and run go test ./provider/k8s (new unit tests cover the ConfigMap behavior).
2. After a cleanup run, inspect ConfigMap convox-release-cleanup in the system namespace; its last-release-build-cleanup key now updates instead of the namespace annotation being rewritten.

### Checklist

- [x] New coverage tests
- [x] Unit tests passing (go test ./provider/k8s)
- [ ] E2E tests passing
- [ ] Documentation updated


ndbroadbent:
Sorry I got bad information from ChatGPT. This change wasn't necessary at all, we can just do this:

```
lifecycle {
  ignore_changes = [
    metadata[0].annotations["convox.io/last-release-build-cleanup"]
  ]
}
```

PR #1059, opened 2026-07-16
guard builds against dropping running-release env

ntner:
### What is the feature/update/fix?

**Update: Build Guard Against Dropping Running-Release Environment Variables**

The CLI now runs a preflight check when you create a build with `convox build`, `convox deploy`, or `convox test`. New builds inherit their environment from the app's newest release, which can differ from the currently running release when environment changes are staged with `convox env set` or `convox env unset` without `--promote`. If the release a build is about to inherit from is missing environment variables that are set in the running release, the CLI now stops before building and explains exactly which variables would be dropped and how to keep them.

The safeguard lives in the shared build path, so it protects `convox build`, `convox deploy`, and `convox test` from a single place, including external builds. When it triggers, the output looks like this:

    ERROR: this build will drop env var(s) that are set in your running release RABC123: SECRET_KEY

    These vars are present in the running release (RABC123) but missing from the latest release (RDEF456), which is what a new build inherits from. This usually happens after `convox env set` or `convox env unset` without --promote.

    To keep them, set them again with --promote before deploying, for example:
        convox env set SECRET_KEY=... --promote

    If you meant to drop these vars, or you believe this is a false alarm, re-run with --force

The check is precise and conservative. It only triggers when the newest release differs from the running release and one or more keys set in the running release are absent from the newest release. Adding new variables or changing values never triggers it, so `convox env set NEW=value` followed by `convox deploy`, and stage-then-deploy workflows, behave exactly as before. The check is also fail-open: if the app has no promoted release, a lookup fails, or the release list is momentarily stale, the build proceeds normally. The safeguard never blocks a build because of its own lookup problems.

---

### Why is this important?

Environment variables are part of your app's running configuration, and changes to them should be visible and deliberate. This guard makes any pending environment drop visible at the moment it matters, before a new release is built and promoted, so the variables can be restored with `--promote` before anything changes.

**Benefits:**

- **Protects running apps during deploys.** The CLI surfaces any pending environment drop up front, with the exact variable names, before a new release is built and promoted.
- **Actionable guidance.** The message names the affected variables, identifies both releases being compared, and shows the exact `convox env set KEY=... --promote` command to keep them.
- **Consistent coverage.** One check covers `convox build`, `convox deploy`, and `convox test`, so every path that creates a build gets the same protection.
- **Never gets in the way of normal workflows.** Additions and value changes pass through untouched, lookup failures fail open, and `--force` remains available as an escape hatch if the guard misfires or the drop is intentional.
- **Works everywhere.** The safeguard is implemented entirely in the CLI using read-only API calls, so it works with apps on every rack version and every provider, with no rack-side or API changes.

---

### How to use it?

No action is needed for normal workflows. The check runs automatically on every `convox build`, `convox deploy`, and `convox test`.

If the safeguard triggers and you want to keep the variables, set them again with `--promote` before deploying:

    $ convox env set SECRET_KEY=value --promote -a my-app

If the drop is intentional, or you believe the guard is wrong, `--force` bypasses it for that command:

    $ convox deploy --force -a my-app

    $ convox build --force -a my-app

With `--force`, the build proceeds and the CLI prints a warning noting which variables are being dropped:

    WARNING: proceeding with --force; this build drops env var(s) set in the running release: SECRET_KEY

#### New CLI Flags

| Command | Flag | Effect |
|---------|------|--------|
| `convox build` | `--force` | Escape hatch: proceed even when the guard detects a pending environment drop |
| `convox test` | `--force` | Same as `convox build` |

`convox deploy` already accepts `--force`. On deploy, the flag now also bypasses this safeguard in addition to its existing meaning of forcing the promote without waiting for readiness.

---

### Does it have a breaking change?

**No breaking changes** are introduced with this update, though there is one intentional behavior change to be aware of. A `convox build`, `convox deploy`, or `convox test` that would produce a release missing environment variables set in the running release now stops and reports the pending drop, including cases where the variables are optional and the build would otherwise have completed. This is deliberate: a pending configuration change to a running app should be confirmed, not applied silently. The guard can be bypassed per command with `--force`, and workflows that only add variables or change values are completely unaffected.

The change is CLI-only and additive. No API endpoints, SDK methods, response structs, or rack-side behavior are modified, and all existing flags and output formats are preserved.

---

### Requirements

This update requires version `3.25.1` or later for the rack.

**Update the Rack:** Run `convox rack update 3.25.1 -r rackName` to update to this version.

The safeguard itself runs in the Convox CLI, so updating your CLI to `3.25.1` enables it for apps on any rack version.

_Note that your rack must already be on at least version `3.24.0` before performing this update._

_If you're unfamiliar with v3 rack versioning, we recommend reviewing the documentation on [Updating a Rack](https://docs.convox.com/management/cli-rack-management/) before applying any updates._

### DIFF
--- a/pkg/cli/builds.go
+++ b/pkg/cli/builds.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -77,7 +78,7 @@ func resolveSrcCredsPass(c *stdcli.Context) (*string, error) {
 
 func init() {
 	register("build", "create a build", Build, stdcli.CommandOptions{
-		Flags:    append(stdcli.OptionFlags(structs.BuildCreateOptions{}), flagRack, flagApp, flagId),
+		Flags:    append(stdcli.OptionFlags(structs.BuildCreateOptions{}), flagRack, flagApp, flagId, flagForce),
 		Usage:    "[dir]",
 		Validate: stdcli.ArgsMax(1),
 	}, WithCloud())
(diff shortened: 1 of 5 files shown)


radekl:
that is quite a breaking change for our CI/CD workflows 😢 

Context - we do utilize this to set environments first through CICD and then we deploy a new version which will use the new variables (that may be not available or break the older version).

I have pushed the PR to fix the CICD at least: https://github.com/convox/action-deploy/pull/15

